PR #140, opened 2015-09-12
Update MapGenScatteredFeatureModBiomes.java

Ezoteric:
This so far fixed it for me, i still need to find a desert to see if it spawns a desert, but in the same world i have a jungle temple and swamphut and no crash. Im still flying arround :P


whichonespink44:
Nice, I've merged it to master and will start flying around myself shortly :)

I guess this means it's possible that more than one scattered feature could generate in the same area?

PR #142, opened 2015-09-16
Update SupportBOP.java

Ezoteric:
Added missing biomes and type support (for temples for example) see what you do with it ;)

### DIFF
--- a/src/main/java/rtg/support/SupportBOP.java
+++ b/src/main/java/rtg/support/SupportBOP.java
@@ -2,6 +2,8 @@ package rtg.support;
 
 import net.minecraft.block.Block;
 import net.minecraft.init.Blocks;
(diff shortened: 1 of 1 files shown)


Ezoteric:
Btw i tested this.

Here is a map without desert stuff but i have found swamp huts in 2 bop biomes :P 
im still rendering maps but my code it awesome :P

Ezoteric:
Ow yeah, i dunno why but my bop doesnt listin to bops biome config when i disable them, can we make a rtg-bop.cfg for that ?

whichonespink44:
I couldn't wait to get home so I could have a proper look at this!! :) I've just merged your stuff now, so I'll give it a test drive in a bit.

As for the BOP config thing... yep, I'll add that to the issue tracker.

Nice work, mate!! :+1:

PR #167, opened 2015-09-22
Rwg base biome removal

whichonespink44:
Removed legacy 'base' biomes & legacy 'realistic' biomes.


Ezoteric:
Btw about gen layers it has a file called mutation I dunno but it sounds like a bleeding thing so we might want to skip that also .

Tbh I think this layer Shit is gonna make you crazy haha

PR #183, opened 2015-09-25
Update ConfigBOP.java

Ezoteric:
Now its correct ;)

### DIFF
--- a/src/main/java/rtg/config/ConfigBOP.java
+++ b/src/main/java/rtg/config/ConfigBOP.java
@@ -124,7 +124,7 @@ public class ConfigBOP
 			generateBOPcrag = config.getBoolean("generateBOPcrag", "BOP Biomes", true, "");
 			generateBOPdeadForest = config.getBoolean("generateBOPdeadForest", "BOP Biomes", true, "");
 			generateBOPdeadSwamp = config.getBoolean("generateBOPdeadSwamp", "BOP Biomes", true, "");
-			generateBOPdeciduousForest = config.getBoolean("generateBOPdeciduousForest*", "BOP Biomes", true, "");
+			generateBOPdeciduousForest = config.getBoolean("generateBOPdeciduousForest", "BOP Biomes", true, "");
 			generateBOPdenseForest = config.getBoolean("generateBOPdenseForest", "BOP Biomes", true, "");
 			generateBOPeucalyptusForest = config.getBoolean("generateBOPeucalyptusForest", "BOP Biomes", true, "");
 			generateBOPfen = config.getBoolean("generateBOPfen", "BOP Biomes", true, "");


whichonespink44:
Sneaky little asterisks!! :snake:

Ezoteric:
Haha BTW I'm checking this latest build of yours and see what's gonna happen haha

whichonespink44:
Hopefully the same thing that happen with 0.0.6!!!  :wink:

Ezoteric:
well, i still get a standard vanilla biome, and also i hardly get any ocean...

The only ocean i have is in the top right corner and all maps are like this. only 1 ocean and rivers between biomes are mostly treeless :P

rtgtest

whichonespink44:
I wonder if it has something to do with the temp/rain values we're using in BiomeBase.java? Those were the setting from RWG, which might've been perfect for the old base biomes, but maybe we need to play around with some different values for those? Or maybe add some additional climates... like maybe:

Very Hot
Hot
Warm
Normal
Cool
Cold
Very Cold

Just thinking maybe that will be enough to smooth out the transitions a bit? I wish I had a super computer so I could load a new world & generate a map in less than a minute!!!  :sob:

Ezoteric:
lol in less then a minute haha keep dreaming :P

And yes, that was what i said before the more the better transition, but it seems RWGs transition code isnt really working or does it ?

I mean in chunkmanager file you have the hot cold wet shit but i think thats not working like it should.

Ezoteric:
Here i created a gist:

https://gist.github.com/Ezoteric/b05fa6529f4afcfd4151

See if you like it ;)

Ezoteric:
btw you deleted the volcano map :p

Ezoteric:
Hmmz, i think i found something important. 
I think rtg/rwg skips genlayer from MC therefor were not getting islands for 1 thing to name.

I was looking into some files and EB BOP etc all use genlayer.
Now if you look at GenLayerBiome.java from MC and biomemanger from forge youll see we need to stop using RWGs way and update the layer files with our own stuff.

I think THAT is the way to get what we want, cause MC self allready has code to make sure desert aint bordering ice biomes etc

whichonespink44:
_"I think THAT is the way to get what we want, cause MC self allready has code to make sure desert aint bordering ice biomes etc"_

It's worth a shot... if we could still generate breathtaking, realistic terrain whilst doing pretty much everything else the way vanilla MC does it, that would be perfect. After I'm done refactoring the two 'Chunk' files in the `truffle-shuffle` branch, that's the last of the rewrite that I'm doing, so we'll be able to start working on the GenLayer stuff. You still up for that screenshare session to get GD up and running on your system?

whichonespink44:
_"btw you deleted the volcano map :p"_

It will be back :wink:

Ezoteric:
Yes I'm still up for that even though I never used hangouts :p

Ezoteric:
Btw at what time you want to do that

whichonespink44:
I can do anytime from now until 10pm GMT... instead of Hangouts then, let's do a Join.Me meeting... you'll need to download and install the client, but after that's it's pretty straightforward. Just let me know what time's best for you and I'll PM you link via Minecraft Forums.

Ezoteric:
now is fine by me pm me your name on MCF

PR #191, opened 2015-09-26
Close, but not close enough

whichonespink44:
I've done what I can, but in order to get GenLayers working as they should, I think we need to make RTG realistic biomes extend BiomeGenBase... which isn't going to be easy.


Ezoteric:
Yeah, thats what i was thinking some time ago, but i thougt it allready did somehow haha

btw, i just saw EB also changes village stuff....
Is this true ? cause if so EB is fuckin DOPE !!!!!!
its one of the things i want in RWG hahah different houses and shit.

Anyway, maybe before we go crazy here, and island is like one of the things we do this, maybe look at how EB does it, or ask in forums if there is a biome way to add islands like how we thougt it could be done, like give tropical island the hardcode it can only spawn in or next to deepocean and deep ocean only to ocean. if we do this we might can skip genlayers

Ezoteric:
BTW, cant we extend our biomebase to biomegenbase ? :P

Ezoteric:
also eb has a biomebase extend but its quite empty sadly :P
But cant we extend it just for the changes it needs ?

Ezoteric:
well, i dunno anymore, if we cant use layers, we need to use the code we wanted to try in from the start. Just let me know what you want to do if this layer isnt gonna happen :)

PR #193, opened 2015-09-26
More re-factoring and experimental tweaks to biomes

whichonespink44:


Ezoteric:
Doezelen the mushroomisland work like this?  :p

Ezoteric:
Does... lol

whichonespink44:
Nope, it still generates inland :cry:  ...  but I'm going to try a few other things today before going back to GenLayers.

Ezoteric:
aaawh :P

well i think i gonna try adding all EB stuff to this cause EB is the BOMB !! :P

whichonespink44:
I got really close with EB's gen layer stuff, but after I got it all setup, it kept crashing at the same spot as the BOP stuff did.... I think my next attempt will involve getting just an empty genlayer system in place that doesn't do anything, and then I'll start adding in functionality piece by piece until it breaks.

Before that though, I need to release 0.0.7 since the debug code I wrote isn't working on servers. Should have it released in the next hour.

Ezoteric:
Ow weird. Do you still have that eblayer setup ? Saves me doing the hassle :p haha

Yeah I saw on mcf :)

Ezoteric:
Maybe the problem is that MC has too many ways to get the vanilla biomes orso and we need to make sure it sees our as their ?

PR #220, opened 2015-10-11
Various bug fixes.

whichonespink44:


Ezoteric:
What does the new biome code do?

whichonespink44:
It's just a basic way to handle when no biomes have been added to a biome category. Previously it would crash if you disabled all COLD biomes for example. But now, if no biomes have been found in a particular category, it tries to grab one from another category, and if it still can't find one, it just grabs any random biome that's been added.

PR #246, opened 2015-10-27
Add a basic build.gradle

Baughn:
Builds as-is. I haven't tested if it runs, but probably.

Well, I mean, the server does.


whichonespink44:
Thanks @Baughn - hopefully this will make it a little more inviting for people to test dev versions :+1:

PR #262, opened 2015-11-03
Fixed issue with scattered features generating everywhere.

whichonespink44:


Ezoteric:
What exactly is everywhere ? Haha

whichonespink44:
Places where they shouldn't ;)

Ezoteric:
Very smart answer. ... I mean did they spawn in wrong biomes or wrong places in the right biome. :p

Ezoteric:
Because the jungle temple often spawn in the middle of the lake hahaha that's also a wrong place so I would suggest if possible to not let the pyramid and temple spawn on water ;)

whichonespink44:
_"I mean did they spawn in wrong biomes or wrong places in the right biome."_

Wrong biomes AND wrong places... but both should be fixed in 0.0.15

BOP registers some biomes oddly, which is why they might seem to be appearing the wrong biome... Steppe, for example, is registered as a SANDY biome, so Desert Temples can gen there.

Ezoteric:
Hmmz didn't I told you this way back when we started ? Or I might explained this to Jimmy hahaha anyway yeah I know a lot of bop biomes will have temples not just desert haha but most where good. ;) but I last copley of months I didn't test with other biome mods so thats why I asked as the only problem I found was the water Shit cause the big lakes are part of the biome so temples will spawn there. Get the point ? ;)

whichonespink44:
Yep, got it. I haven't seen any structures generating on water, or anywhere they're not supposed to using the latest version... if you look in the console, you can see that the game is still **trying** to gen them where they don't belong (I'm logging every attempt in the console), but I think I've fixed it for good now, so they will only spawn where they're allowed to... the only exception are buried desert temples, but that's hardcoded in Minecraft and not sure I want to fix that.

I could be talking out of my ass though, so do let me know if you find any structures spawning where they're not supposed to.

Cronosus:
and what about this? :D

PR #297, opened 2015-11-27
Fix wrong encoding in .lang files

Adaptivity:
Converted UTF-8 BOM encoding to UTF-8 (Minecraft can't read UTF-8 BOM encoding)
Changed Chinese localization from "真实世界生成初版" (which Google translates as "Real world generation first edition (alpha version)) to "真实世界生成" which means just "Real world generation".


whichonespink44:
@Adaptivity Cheers mate!

PR #311, opened 2015-12-03
Re-decorated vanilla Jungle biomes.

whichonespink44:


Ezoteric:
HEY !!! i forgot to tell you this...

but when i enter a jungle biome (in the latest release) my FPS go down to 1 !!! 

Now i havent tested this withtout mods but it was full of trees hahaha

whichonespink44:
The vanilla jungle biomes in 0.1.0 are just using the vanilla decorations. It's still RTG terrain, but the decorations (trees, bushes, etc) are just plain vanilla.

I've actually re-decorated them in the latest dev version, so maybe give that a try and let me know if it's any better. Although, jungles are always going to be the laggiest biomes because of all the block updates (vines, cocoa plants, decaying leaves, etc.) Not much I can do about that unfortunately. :(

Ezoteric:
Hmmz I didn't know that. Man mc started to get leggy since 1.7.10 with 1.6 I could have twice as more mods and no lag and now I have half he mods and need to turn off a lot of animations or I can't play at all :( stupid mc :p  

But my map is fucking awesome. A lot of oceans but still a bit sad these are biomes but at least I got a few big ones :) 

I'll upload the picture tomorrow ;)

PR #349, opened 2015-12-18
fixed max height

Ezoteric:
ice mountains should top at max y-200 now


whichonespink44:
Closing this as it's targeting the `master` branch. All future pull requests should target `dev`. (My bad for not mentioning it.) I've added the change manually though, so thanks for that!! :smile:

PR #376, opened 2016-01-05
Upped the chance to spawn palm on beach

Ezoteric:
As the beach is so damn small now it now creates at least a few palms so you know and get the feeling its a beach :P


whichonespink44:
This is the line you should be tweaking to increase the spawn rate:

`if (rand.nextInt((int) (4f / strength)) == 0) {`

Try changing 4f to 3f (or 2f) to see what happens.

Alternatively, you can replace that line with:

`if (rand.nextInt(4) == 0) {`

to make them have a 25% chance. 5 = 20% chance. 3 = 33% chance. 2 = 50% chance

PR #383, opened 2016-01-06
Alternate dimension fix and *critical* fix to default worlds.

Zeno410:
Only provides RTG chunk managers to the overworld. Also doesn't do RTG generation changes to non-RTG worlds.


Zeno410:
Left in some testing code

PR #384, opened 2016-01-06
No RTG chunk manager for alternate dimension and *critical* fix to OnBiomeGenInit

Zeno410:
As before, but without the testing line


Zeno410:
Never mind; just spotted a mistake

PR #397, opened 2016-01-09
Coastal Biome Repair procedures

Zeno410:
Also a lot of bogus commits consisting of attempts to update from the master. Currently the correct biome seems to be doing the surface but somehow sometimes the uncorrected biome ends up on the F3. Underwater is not fixed - there are still land biomes underwater and there's a lot of river that shouldn't be river.


whichonespink44:
1,054 files changed, due to file mode changing.

I'm reluctant to add this to the `master` branch in case it has a knock-on effect on things we might not be aware of yet.

Any chance you could submit the pull request again, except target the [`dev`](https://github.com/Team-RTG/Realistic-Terrain-Generation/tree/dev) branch instead?

Ezoteric:
guess hes sleepin

whichonespink44:
@Zeno410 Can you try running the 2nd command from [this article](https://github.com/Gizra/KnowledgeBase/wiki/How-to-set-Git-to-ignore-chmod-changes) and submitting the pull request again?

PR #405, opened 2016-01-12
Compatibility with biome limit extending mods.

elix-x:
Biome id limit of 256 can be extended using a lot of asm. But it is possible.
Mods such AIC 2.0 and NEID have done it, so to support them you just have to make your biome array have length of vanilla's.

### DIFF
--- a/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
+++ b/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
@@ -41,7 +41,7 @@ import net.minecraftforge.event.terraingen.TerrainGen;
 
 public class RealisticBiomeBase extends BiomeBase {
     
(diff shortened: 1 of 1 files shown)


Zeno410:
This looks straightforward, but there's more code than this that will need to change. The biome repair code will also need fixing, and probably some other things as well. But we should still take the pull.

whichonespink44:
Thanks for that @elix-x :+1:

Zeno410:
I searched the code base for "256", many of which references are to the biome array and need a similar change:

Found 4,914 match(es) in 352 file(s)

Um, yes, a bit more work! The vast majority are references to the 256 height but somebody's gotta look at a lot of code.

whichonespink44:
I can do the refactoring. @Zeno410 does it need to be done before we release snapshot 5? either way, I'll create a new issue for this.

elix-x:
Awesome!
Now i just have to report the same thing to BoP, highlands, extra biomes...

Issue #391, opened 2016-01-07
Snapshot 5 - problem with biome 161


KeepOnDigging:
I am testing with snapshot-3. 
Climate Control is not installed. 
B:"RTG Biome Layout"=false
I have no extra biome mods installed. 
I'm just using RTG to create a nicer-looking vanilla world.

I have tested this with only RTG installed, no other mods.

If I create a new realistic world with seed: 32567400, Minecraft crashes immediately.
I think it also crashes with any other seed when biome 161 is first encountered.

This may (or may not) be related to the incorrect name for biome 161 in Minecraft-1.7.10. This biome name is missing the "M" so the name is a duplicate of biome 160. Biome dump from NEI shows:

```
160,Mega Spruce Taiga,0.25,0.8,0.1,0.2,0.2,"DENSE, FOREST",net.minecraft.world.biome.BiomeGenTaiga
161,Mega Spruce Taiga,0.25,0.8,0.1,0.2,0.2,"DENSE, FOREST",net.minecraft.world.biome.BiomeGenTaiga
```

The relevant part of the crash log:
```
[16:21:03] [Server thread/ERROR]: Encountered an unexpected exception
net.minecraft.util.ReportedException: Exception generating new chunk
	at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:631) ~[MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:547) ~[MinecraftServer.class:?]
	at net.minecraft.server.integrated.IntegratedServer.func_71217_p(IntegratedServer.java:111) ~[bsx.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:427) [MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer$2.run(MinecraftServer.java:685) [?:?]
Caused by: java.lang.RuntimeException: problem with biome 161
	at rtg.world.biome.WorldChunkManagerRTG.func_76936_a(WorldChunkManagerRTG.java:103) ~[WorldChunkManagerRTG.class:?]
	at net.minecraft.world.biome.BiomeCache$Block.<init>(SourceFile:26) ~[ahz.class:?]
	at net.minecraft.world.biome.BiomeCache.func_76840_a(SourceFile:52) ~[ahy.class:?]
[...]

PR #406, opened 2016-01-13
Fix for issue #391, biome 161

KeepOnDigging:
Add a work around for Minecraft bug resulting in missing vanilla biome id 161. This fixes issue #391.


whichonespink44:
Thanks @KeepOnDigging :+1:

PR #411, opened 2016-01-13
Made Fens into swamps

Zeno410:
will stop beach conversion of Fen biome


Ezoteric:
Yeah is a swamp I checked ebs biome types

Issue #408, opened 2016-01-13
Refactor all hardcoded instances of the biome array length


whichonespink44:
From #405:

_"I searched the code base for "256", many of which references are to the biome array and need a similar change:

Found 4,914 match(es) in 352 file(s)

Um, yes, a bit more work! The vast majority are references to the 256 height but somebody's gotta look at a lot of code."_

Zeno410:
No urgency on this one. As long as nothing's trying to use extended biome tables either the old 256 approach or the new "size of biome array" will work just fine. We just have to have EVERYTHING fixed for extended biome tables to work properly.

topisani:
This is in v2. Do we need it here too?

whichonespink44:
Not sure.

@elix-x Are there any mods taking advantage of the AIC's extended biome array length in 1.7.10? If not, I'd be a little reluctant to proceed with this as I don't want to run the risk of introducing any new bugs. In theory, everything should be fine, but... it just feels like we could be opening up a can of worms (bugs) by implementing this now.

elix-x:
Not yet. Meanwhile, i recently finalized API, so i'll be pulling others into this mess.
Considering i pulled you in first, and some time after you moved to climate control based gen, recently i  pulled Zeno in. So, while climate control isn't adapted with AIC, well, RTG cannot be fully used.

So, it's up to you know.

BTW, i still haven't gotten any news from Zeno, so @Zeno410 ???

whichonespink44:
@elix-x Do you have plans to update AIC for 1.9? If so, I think we should try to add support for this in our 1.9 version instead, since we're already in the middle of a [major re-write](https://github.com/Team-RTG/1.9-Realistic-Terrain-Generation/tree/RTG2) of RTG for 1.9.

elix-x:
I saw that you're working on 1.9, and i have plans on AIC for 1.9, just not yet (i don't thinkg enough biome mods are there for me to move, meanwhile 1.7.10 is getting old). Although, API class will be the same (just different imports related to fml).

srs-bsns:
@elix-x, right now the only mods in 1.9 that add biomes, that I know of, are BoP and AbyssalCraft. As for the other major mods that add a lot of biomes, there might not be many that end up getting ported to 1.9.

1. Highlands, ~40 biomes (Zeno was given permission to take this over, so a future port is expected, eventually)
[...]

PR #464, opened 2016-01-31
AIC 2.0 Compatibility

elix-x:
These are all the methods that have to be modified to make RTG and AIC
fully compatible.
Note: this part of AIC API can be redistributed freely.

AIC 2.0 is the mod i was talikng earlier, which increases biome ids limit.
This is it's API and necesarry changes to be compatible with RTG.
If you want dev version of AIC, or have any questions, contact me over at minecraft forums (i am [elix_x ](http://www.minecraftforum.net/members/elix_x)over there).

AIC - Anti Id Conflict.

This is 2nd try of me pull requesting. Now i updated my fork, and targeted dev (as you asked).
Also, this part of AIC API can be redistributed and does not contain any big and complex methods and consists of 1 class.

### DIFF
--- a/src/main/java/code/elix_x/coremods/antiidconflict/api/AICChangesWrapper.java
+++ b/src/main/java/code/elix_x/coremods/antiidconflict/api/AICChangesWrapper.java
@@ -1,0 +1,52 @@
+package code.elix_x.coremods.antiidconflict.api;
+
+import cpw.mods.fml.common.Loader;
+import cpw.mods.fml.relauncher.ReflectionHelper;
+import net.minecraft.world.chunk.Chunk;
+
+/**
+ * This class wraps all methods that were changed (name or signature) in vanilla by AIC.
+ * Methods here should be used instead of vanilla's, when AIC is loaded, or otherwise your calls to vanilla methods will be redirected and voided to not cause any crashes.
+ * <br>
+ * This part of AIC API can be redistributed freely, but not modified.
+ * @author elix_x
+ */
+public class AICChangesWrapper {
+
+	/**
+	 * Simple method to check if AIC is loaded.
+	 * @return if AIC is loaded or not.
+	 */
+	public static boolean isAICLoaded(){
+		return Loader.isModLoaded("AIC");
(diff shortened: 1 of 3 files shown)


whichonespink44:
@Zeno410 What are your thoughts on how this has been implemented? We'll still need to refactor all hardcoded instances of the biome array length as per #408, correct?

@elix-x I just noticed this PR is still targeting `master`, but rather than asking you to re-target to `dev`, I'm going to create a feature branch for this. When it's been created, would you mind closing this and doing another PR targeting that branch?

whichonespink44:
Created a new branch from `dev` called `AIC-2.0-Compatiblity` - https://github.com/Team-RTG/Realistic-Terrain-Generation/tree/AIC-2.0-Compatiblity

elix-x:
Yes. I'll do it right now.

elix-x:
Here: https://github.com/Team-RTG/Realistic-Terrain-Generation/pull/470

whichonespink44:
Awesome, thanks! :)

Zeno410:
Comments: 
a) IsAICLoaded should be stored in an instance var for performance, I think.
b) The Forge classloader might get upset loading the method and/or class if AIC is not around. I find I usually have to put references to external mods into wrapper classes. But try it and see.
c) I'm not clear on how a reflected reference to a method in a set class (Chunk.class) will resolve differently from a direct call.

elix-x:
A) Not very much perofmance is in game. But i'll think.
B) No. This class can be freely redistributed and does not depend on any other AIC classes. Class loader will not get upset.
Note: kinda similar thing is done by RF API: it can be redistributed, and classes loaded even without rf mods present.
C) I'm changing method's signatures (getChunkArray and setChunkArray) from ()[B and ([B)V to ()[I and ([I)V (ASM), thus for java, it's a completely different method, and i have to remove all invocations of old ones to avoid MethodNotFound crashes. As Forge does not allow to see changes applied by coremods in workspace, i cannot invoke it directly, as there's not getChunkArray([I)V method.

PR #470, opened 2016-02-02
Aic 2.0 compatiblity

elix-x:
These are all the methods that have to be modified to make RTG and AIC
fully compatible.
Note: this part of AIC API can be redistributed freely.

AIC 2.0 is the mod i was talikng earlier, which increases biome ids limit.
This is it's API and necesarry changes to be compatible with RTG.
If you want dev version of AIC, or have any questions, contact me over at minecraft forums (i am [elix_x](http://www.minecraftforum.net/members/elix_x)over there).

AIC - Anti Id Conflict.

3rd try, targetting AIC-2.0-Compatibility branch.


elix-x:
Do you want me to contact you over at mc forums, and send you dev version of AIC, or no?

whichonespink44:
Yes, that would be great if you could, thanks.

PR #504, opened 2016-02-13
Updated AIC API.

elix-x:
Updated AIC API to support ability to disable biome ids limit extension.
This is "hopefully"-(tm) last time API changes.
But it changes for good.

PS: Wait for alpha 8, to test.


whichonespink44:
@elix-x We've moved the AIC API from src/main to src/api, so you'll need to update your branch before submitting the PR.

PR #505, opened 2016-02-13
Updated AIC API

elix-x:
Updated AIC API to support ability to disable biome ids limit extending.

I finally managed to update my fork (after deleting it 2 times) and rewriting update code (i rewrote it 10 times, because nothing was working, and github reset it and things and stuff).

To test with AIC, wait for alpha 8.


whichonespink44:
So with these latest changes, users will be able to use AIC 2.0 to detect biome ID conflicts and optionally they will be able to use biome IDs greater than 255, correct?

elix-x:
Exactly. But id limit extension is activated by default.

PR #535, opened 2016-02-26
Fixed bug where Natura only spawns blueberries.

whichonespink44:
Now manually setting temp/rain values in BiomeBase. Thanks @srs-bsns !!


srs-bsns:
Works good. :+1:

PR #558, opened 2016-03-07
Biome Repair Smoothing

Zeno410:
Smoothing features added to biome repair. Beach-land mismatches mostly
eliminated; diagonal jagged chunk borders fixed; some reduction in
straight-line artifacts

### DIFF
--- a/src/main/java/rtg/util/CircularSearchCreator.java
+++ b/src/main/java/rtg/util/CircularSearchCreator.java
@@ -6,20 +6,33 @@ package rtg.util;
  * @author Zeno410
  */
 public class CircularSearchCreator {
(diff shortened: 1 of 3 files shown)


whichonespink44:
Nice one :+1: Can't wait to see it in-game!! :)

Zeno410:
Well, it's not too exciting to look at - a number of problems just don't show up or much less often. The main thing left is the line where beaches end from being too far from the sea. It's fixed if it was a diagonal zaggedness but not if it was straight. I thought I was going to get more improvement on that than I did. Definitely diminishing returns here - this was as much work as the original biome repair but for a lot less benefit.

PR #562, opened 2016-03-09
Terrain revisions for 0.8

Zeno410:
Not all done, even for vanilla, but there is stuff to look at here. Lakes may be too large, but that can be changed and I think they're much better than the old system. One big change is that I added a noLakes variable to RealisticBiomeBase and it needs to get set for any biome that shouldn't have lakes, which is most high and hilly biomes (the banks look bad). I only did that for vanilla.


whichonespink44:
Nice, just giving it a test drive now.

whichonespink44:
Lookin pretty good!

Rivers feel windier, which is great. Did you also increase the average width, or did I just get lucky with this seed?

Couple of questions regarding the lakes...

1) Are the lakes with sandy shores the only lakes generated with the new system?

2) Do we have any control over the size of the lakes on a per-biome basis, or would that be more of a general thing?

3) Is it possible to have gravel shores, or even shoreless lakes?

Zeno410:
I don't think I did anything to river courses. They're a little deeper and wider. The main reason is that I'm using the same code for rivers and lakes and the lakes were too shallow. The rivers probably should have been a bit deeper anyway. 

The lakes generate without any shore. A number of biomes place beach materials (sand and gravel) at beach level but the biome stays the land biome. The biome repair will also affect lakes near the ocean, turning the underwater sections into Ocean biome and appropriate heights into genuine Beach/Stone Beach/Cold Beach. At present it can only do that near the ocean. It could probably be rewritten to do the same inland although it's not trivial, especially since I don't want to do that on rivers.

Some biomes do leave gravel shores, and some I think are shoreless. Again, it depends on the biome coding.

The best way to check whether something is a lake or an ocean is to check the biome. If you're more than (I think) 64 blocks from "real" ocean lakes carry land biomes *except* if there's supposed to be a river there it becomes River. I actually changed to an all-land world so I didn't have to worry about it.

The code to calculate lake parameters is in RealisticBiomeBase and it could be overridden by the biome to make lakes bigger or smaller (also deeper/shallower and more or less irregular, for that matter). That might come out weird at biome junctions, although the smoothing should stop the worst parts of that. You do occasionally get strange stuff at junctions with lakeless biomes. Just by accident, the test world I'm working in spawns me right next to a MegaTaiga Hills, which doesn't have lakes, and it's *right in the middle* of what's supposed to be a lake. So I end up with a bunch of ponds around the base.

Originally I was planning to override to stop lakes in biomes that shouldn't have them but when I realized how many overrides I'd have to write I switched to the noLakes variable (which could be make configurable).

PR #583, opened 2016-03-21
Dev version of the travis stuf

topisani:


whichonespink44:
Merged master into dev and bumped to 0.8.0, so it should be alright - let me know if not.

Issue #218, opened 2015-10-10
EB-related crash: 'Exception getting block type in world'


whichonespink44:
When running RTG with EB, the game crashes when trying to enter the world for the first time.

**Crash reports:**
https://gist.github.com/whichonespink44/48b0d15d1a6b9d022080
https://gist.github.com/whichonespink44/83e7c9dc131c86661043
https://gist.github.com/whichonespink44/030e74a4d20864e839dc

It seems to be run into trouble when RTG tries to decorate the EB biome:

```java
    public void rDecorate(World world, Random rand, int chunkX, int chunkY, OpenSimplexNoise simplex, CellNoise cell, float strength, float river) {
    
        if (strength > 0.3f) {
            baseBiome.decorate(world, rand, chunkX, chunkY);
        }
    }
```
https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/master/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java#L69

**How to repeat the crash:**
Install RTG ([0.0.10](https://github.com/Team-RTG/Realistic-Terrain-Generation/releases/download/v0.0.10/RTG-1.7.10-0.0.10.jar)) & EB (2.5) with no other mods and try to start a new world.

whichonespink44:
@SMEZ1234 - Sorry to trouble you... but is there anything from these crash reports that immediately jumps out at you as being the potential culprit? I'm hoping it's something really simple that we've overlooked when adding support for EB, but after several of debugging this over the last couple of days, I haven't made much progress unfortunately.

**Other potentially useful information:**

* I've tried disabling the EB biomes from the stack traces, but it seems to happen with every biome (at least, all the ones I've tested).

* I've tried dozens of different combinations of config settings (EB ores disabled, EB stone disabled, vanilla module disabled, etc), but it's the same result each time.

* Here's where all of the EB biomes are located: https://github.com/Team-RTG/Realistic-Terrain-Generation/tree/master/src/main/java/rtg/world/biome/realistic/enhancedbiomes

* And here's where they're being added during postInit(): https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/master/src/main/java/rtg/world/biome/realistic/enhancedbiomes/RealisticBiomeEBBase.java#L22

* I've added support for the other biome-adding mods in more or less the same way, and they all seem to be fine.


Any ideas?

whichonespink44:
[...]

PR #598, opened 2016-03-27
Assorted fixes

Zeno410:
Fixes to high river banks, inappropriate river canyons, Kelp Forest and Coral Reef too high, and some BoP biomes.

Designed terrains for some EB biomes; not yet tested.


whichonespink44:
Gonna take a look at the new EB designs now.

whichonespink44:
Getting a crash on world load related to the JitteredEffect:

http://pastebin.com/Xiu6hr63

whichonespink44:
Might be a problem with a particular EB biome because I just tried loading another world, and it hasn't crashed yet.

The seed from the crash report above is:

-8849954191548828809

whichonespink44:
I've been going through the various EB biomes that changed in this PR, and so far the Badlands biome is the only one that's crashing.

However, I've noticed a few things as I've been testing the other biomes...

First, the generation is extremely slow. Sometimes chunks take 30-60 seconds to load in, even at render distance 4, and other times they just don't seem to load at all.

Another thing I noticed is EB's Boreal Plateau biome:



Not sure what's going on there, but all of the water you see is 'River' and all of the land is 'Boreal Plateau'

(It's probably worth mentioning that I'm using Lonely Biome to generate single-biome worlds, so not sure if that has something to do with it. I'm thinking not because the other EB biomes seem to be generating fine.)

I'll keep working my way through the other EB biomes that have changed and report back with any relevant findings.

whichonespink44:
Loving the basalt/gabbro combo with the mesa template for EB Stone Canyon biomes, but the surfacing needs an overhaul as it's making a mess of the cliffs.

All of the other EB biomes seem to be generating fine.

Finally, just some general comments on EB biome design for reference:

1. The EB stones used in each biome were taken directly from 'naturally-occurring' EB biomes, so wherever possible I'd like to keep those because if we change too many, then some stone variants might not be obtainable for EB users. Some stones are used all over the place, but some are only used in a few biomes, and if we replace the rare ones, then they won't get generated at all.

2. Some EB biomes can't be decorated by the base biome (see [this issue](https://github.com/Team-RTG/Realistic-Terrain-Generation/issues/218) for details), so just keep that in mind when re-designing these. If anyone needs to remove decorations, just add a 'dummy' deco object with allow=false and that should prevent the generator from allowing the base biome to decorate itself.

srs-bsns:
EB Badlands is one of my least liked biomes, so I almost don't even care that it's broke. :grinning: We can't have that crashing though.
If I have some free time later (if I take a break from GUI Configs) I'll give this version a test drive.

Badlands: https://www.youtube.com/watch?v=RC9ljHl__pI

whichonespink44:
Haha, love me some Mitch Hedberg :smile: 

The screenshot in that video is a nice interpretation of badlands - wouldn't mind seeing something like that.

image1

Zeno410:
Have a fix in for the EBBadlands. I haven't really been able to test the biomes much - I was trying to get a fix in for Coral Reef/Kelp Forest and a few other river-related things and the EB biomes came along for the ride.

Yes, the surfacing is a mess but I didn't want to get into that - the terrain will keep me quite busy enough for a while.

Which EB Biomes were really slow? Or is it just intermittent (I see that whenever I don't have code looking for slowdowns - seriously!)

Lonely Biome shouldn't have any effect.

PR #605, opened 2016-03-31
Vanilla biome redecoration.

whichonespink44:


whichonespink44:
Some of the vanilla biomes still need some 'decorative' work, but... I wanna get this into dev so I can have access to the enhanced Decos for the new EB biomes.

PR #625, opened 2016-04-10
Lake and mountain adjustments, and river/lake configs

Zeno410:


whichonespink44:
Can't wait to give this a test drive - so many water options!! :)

PR #629, opened 2016-04-13
Optimised imports, Removed static import, Fixed some OCD stuff

srs-bsns:
This should fix your Eclipse problems.

### DIFF
--- a/src/main/java/rtg/RTG.java
+++ b/src/main/java/rtg/RTG.java
@@ -1,7 +1,12 @@
 package rtg;
 
-import java.util.ArrayList;
-
+import cpw.mods.fml.common.Mod;
+import cpw.mods.fml.common.Mod.EventHandler;
+import cpw.mods.fml.common.Mod.Instance;
+import cpw.mods.fml.common.SidedProxy;
+import cpw.mods.fml.common.event.*;
+import cpw.mods.fml.relauncher.Side;
+import net.minecraftforge.common.MinecraftForge;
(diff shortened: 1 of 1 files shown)


topisani:
why would removing static imports fix things?

srs-bsns:
I dunno. Zeno and Pink were both having issues with Eclipse being problematic with the static import, and were having problems compiling the mod inside of the IDE. We should probably try to avoid static imports when neccessary though. [Even the Oracle docs say to avoid it when not neccessary.](https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html)

whichonespink44:
Cool - I'll give it a test drive tonight and see if it works. Nice one! :+1:

whichonespink44:
Still not working for me: http://pastebin.com/M43Ap3XT

Looks like it doesn't like the gradle variable replacement stuff?

srs-bsns:
So, you're not building with gradle in Eclipse?

Even inside of Eclipse, you should be able to set up a run configuration that invokes the gradle script to compile the mod. (Tasks -> forgegradle -> runClient) This is how it's done in IDEA, and even if I don't want to go through the task of manually setting up the run configuration, all I have to do is manually run the gradle task once and it automatically sets it up for me, so the second time I just need to click the 'Play' button.

Srsly, you should switch to IDEA. :relaxed: :smug:

topisani:
everyone should

PR #637, opened 2016-04-15
Biome-specific cave settings.

whichonespink44:
This is what I've done for the biome-specific cave settings.

It looks worse than it was (most of the changed files are the updated example configs). The only two files involved here are:

```
src/main/java/rtg/api/biome/BiomeConfig.java
src/main/java/rtg/world/gen/MapGenCavesRTG.java
```

I've tested it on a few biomes, and everything seems to be working correctly.

The only concern I have is the way I'm checking for the current biome in MapGenCavesRTG:

`BiomeGenBase biome = world.getBiomeGenForCoords(this.rand.nextInt(16) + chunkX * 16, this.rand.nextInt(16) + chunkZ * 16);`

I'm not very good with the whole chunk/world coord stuff, but it seems to work alright.

Can anyone see any problems with how I'm getting the current biome?

### DIFF
--- a/etc/config/RTG/biomes/abyssalcraft.cfg
+++ b/etc/config/RTG/biomes/abyssalcraft.cfg
@@ -32,6 +32,12 @@ biome {
             #  [default: ]
             S:"RTG Surface: Top Block Meta"=
 
+            # This setting controls the number of ravines that generate.
+            # LOWER values = MORE ravines & MORE lag. (50 = vanilla ravine frequency)
+            # Set to -1 to use global setting. Set to 0 to disable ravines for this biome.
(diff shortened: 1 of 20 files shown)


Zeno410:
You need to either create a Random from a passed seed or reinitialize it to something parameter-dependent or you'll get inconsistent results from run to run. In practical terms, it's probably easiest to make a new Random off some hash-like function of chunkX and chunkZ.

whichonespink44:
> create a Random from a passed seed or reinitialize it to something parameter-dependent

I was just going to do this, but apparently MapGenBase already creates a random from the world seed, which MapGenCavesRTG is using, so... I think we're all good.

PR #662, opened 2016-04-23
Highlands re-decoration.

whichonespink44:
I'm still not finished re-decorating the Highlands biomes, but I thought I'd create the PR now in case anyone wants to discuss anything since it contains a lot of changes that will eventually end up in `dev`.

The most obvious change is the addition of the Highlands code to src/api, but there's also the foundation for the new tree system, which I'll be writing some documentation for when I can so I can put it on the wiki.


Zeno410:
`int intX = chunkX + rand.nextInt(16) + 8;
 int intZ = chunkY + rand.nextInt(16) + 8;`

Doesn't that push the trees into adjacent chunks 3/4 of the time? That's highly undesirable for performance.

Also, wouldn't this be a good time to switch to an abstract extensible class rather than an enum? If you don't want to pass all the variables, you could create the tree classes as inner classes in the various mods. External classes would be nice though as we could swap trees between mods. Since the vars are all public you could pass the DecoTree in the generate() call.

topisani:
+1 on the pushing trees into other chunks thing, we dont want that. About the tree system, it is being revised for v2, and while i think we agreed to keep it like this for v1, subclasses could probably be implemented without changing anything at all.

whichonespink44:
> Also, wouldn't this be a good time to switch to an abstract extensible class rather than an enum? If you don't want to pass all the variables, you could create the tree classes as inner classes in the various mods. External classes would be nice though as we could swap trees between mods. Since the vars are all public you could pass the DecoTree in the generate() call.

I'll definitely look into that when the first version of the new tree system is finished as I'll have a much better idea of what's involved.

topisani:
yeah, cause pink can just merge his own PRs... frikin capitalism

whichonespink44:
Down with Thatcher!! :hammer:

PR #667, opened 2016-04-25
Fix for surface ponds

Zeno410:
WorldGenLakes replaced with WorldGenPonds. Ponds are now flush with at
least some land, are now unlikely to leave stuff floating in midair,
and don't intrude into house.

### DIFF
--- a/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
+++ b/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
@@ -35,6 +35,7 @@ import rtg.world.biome.BiomeBase;
 import rtg.world.biome.RTGBiomeProvider;
 import rtg.world.biome.deco.DecoBase;
 import rtg.world.gen.feature.WorldGenClay;
+import rtg.world.gen.feature.WorldGenPond;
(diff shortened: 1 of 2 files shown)


whichonespink44:
1233400426_guy falls of chair

whichonespink44:
I don't think I've ever been more excited to test a PR!! :dancer: 

Does this also fix BOP ponds as well?

Zeno410:
Yes, it does. Bayou, Wetlands, Marsh, etc. - all much better.

topisani:
Y no screenies?

Zeno410:
Lazy. Also, you could get decent ponds from the old system, just not very often, so a screenie doesn't prove much. However, it doesn't seem to be working right in Bayou after all - I'm still getting cutoff trees. It's still a big improvement and I'll take a look at the cutoff trees when I get a chance.

whichonespink44:
Yeah, I've noticed the occasional floating tree or floating piece of tallgrass from where the pond digs into the surface, but... on the whole, it's definitely a massive improvement.

If you think you can improve it even further, go for it... but... even if it doesn't get any better than it is right now, I'd still be happy. I may even enable ponds on my survival worlds now :)

Issue #603, opened 2016-03-29
Performance


whichonespink44:
Not sure what's happened, but generation is being weirdly slow sometimes. It's intermittent for me, but something's definitely going on. Other people have confirmed this in the forums.

We need to get one of those fancy profilers on the scene and see where the bottleneck is.

Zeno410:
I suspect recursive decorations storms, myself. I think Forge has put something in to fix them and sometimes it gets stuck.

whichonespink44:
That would explain it if that's what's happening. Also on the decoration side of things, it occurred to me that each decoration calls its own Forge event, so if there are three tree-related deco objects assigned to a biome, the 'TREE' event gets called 3 times... so that might not be a good thing.

That wouldn't explain the lag with modded biomes though, because I've only re-decorated vanilla, so it might be the terrain and/or river stuff contributing like you mentioned earlier.

srs-bsns:
I noticed some real bad lag today generating vanilla biomes in 1.9 whilst testing AbyssalCraft, consistant with lag I've seen recently in 1.7.10.

To describe it further, it's not a lag where chunks generate slowly but consistantly, it happens in bursts. It will seem like generation has completely locked up, sometimes for 10-15 secs, then the surrounding chunks will 'burst' load, then another freeze for 10-15 secs. It's more noticeable in 1.9 since chunks appear to load differently than in 1.7.10, but it's relatively the same.

I've seen this sort of thing happen when available memory is low and the JVM garbage collector craps out trying to free up memory that it cant, or when the heap becomes badly fragmented, but this isn't the case. This sort of thing happens right from the start, even when generating spawn.

whichonespink44:
Yep, that matches up with what I've been experiencing as well.

Zeno410:
My memory tracker does not show anything unusual when it happens.

Zeno410:
If you'll look at the Forge configs you'll see the likely source of the problem. Forge has a limit on the number of chunks a mod can "force" open. The problem is that the way anything with a lot of big trees works it will "force" open every single chunk in viewing range. I don't know how Forge resolves that issue, but that's your problem - the tree decorators need more chunks and Forge won't let them have it. They have to wait for Forge's asynchronous loader to serve it up.

srs-bsns:
Hrm... I never even thought about that. This is definately worth some experimentation, since mod overrides can be added.

[...]

PR #670, opened 2016-04-28
Symmetrical Dots, WorldGenLakes, Performance and River Noise fixes // all important

Zeno410:
However, wrapped up in some attempts to center decorations on the chunk, so this should probably have some testing first. I think the deco changes have improved things but it could be wishful thinking. I still get generation pauses.

### DIFF
--- a/src/main/java/rtg/util/VoronoiCellOctave.java
+++ b/src/main/java/rtg/util/VoronoiCellOctave.java
@@ -41,11 +41,15 @@ public class VoronoiCellOctave implements CellOctave
 	private boolean useDistance = false;
 
 	private long seed;
+    private long ySeed;
+    private long zSeed;
 	private short distanceMethod;
 
 
 	public VoronoiCellOctave(long seed, short distanceMethod, boolean useDistance){
 		this.seed = seed;
+        ySeed = new Random(seed).nextLong();
+        zSeed = new Random(ySeed).nextLong();
 		this.distanceMethod = distanceMethod;
         this.useDistance = useDistance;
 	}
@@ -109,6 +113,8 @@ public class VoronoiCellOctave implements CellOctave
 	public void setSeed(long seed)
 	{
 		this.seed = seed;
+        ySeed = new Random(seed).nextLong();
+        zSeed = new Random(ySeed).nextLong();
 	}
 
 	public float noise(double x, double z, double frequency)
@@ -130,7 +136,7 @@ public class VoronoiCellOctave implements CellOctave
 			{
 
 				double xPos = xCur + valueNoise2D(xCur, zCur, seed);
-				double zPos = zCur + valueNoise2D(xCur, zCur, new Random(seed).nextLong());
+				double zPos = zCur + valueNoise2D(xCur, zCur, zSeed);
 				double xDist = xPos - x;
 				double zDist = zPos - z;
 				double dist = xDist * xDist + zDist * zDist;
@@ -178,7 +184,7 @@ public class VoronoiCellOctave implements CellOctave
 			for(int xCur = xInt - 2; xCur <= xInt + 2; xCur++)
 			{
 				xPos = xCur + valueNoise2D(xCur, zCur, seed);
-				zPos = zCur + valueNoise2D(xCur, zCur, new Random(seed).nextLong());
+				zPos = zCur + valueNoise2D(xCur, zCur, zSeed);
 				xDist = xPos - x;
 				zDist = zPos - z;
 				dist = distance(xPos - x, zPos - z);
@@ -208,7 +214,8 @@ public class VoronoiCellOctave implements CellOctave
 		dCandidate = dCandidate / total;
 		dNeighbour = dNeighbour / total;
 
-		double c = (diff / 2D) - dCandidate;
+		//double c = (diff / 2D) - dCandidate;
+        double c = (dNeighbour - dCandidate)/dNeighbour;
 		if(c < width)
 		{
 			return (((float)(c / width)) - 1f) * depth;
@@ -239,7 +246,7 @@ public class VoronoiCellOctave implements CellOctave
 			{
 
 				double xPos = xCur + valueNoise2D(xCur, zCur, seed);
-				double zPos = zCur + valueNoise2D(xCur, zCur, new Random(seed).nextLong());
+				double zPos = zCur + valueNoise2D(xCur, zCur, zSeed);
 				double xDist = xPos - x;
 				double zDist = zPos - z;
 				double dist = xDist * xDist + zDist * zDist;
@@ -300,8 +307,8 @@ public class VoronoiCellOctave implements CellOctave
 					// this unit cube.
 
 					double xPos = xCur + valueNoise3D (xCur, yCur, zCur, seed);
-					double yPos = yCur + valueNoise3D (xCur, yCur, zCur, rand.nextLong());
-					double zPos = zCur + valueNoise3D (xCur, yCur, zCur, rand.nextLong());
+					double yPos = yCur + valueNoise3D (xCur, yCur, zCur, ySeed);
+					double zPos = zCur + valueNoise3D (xCur, yCur, zCur, zSeed);
 					double xDist = xPos - x;
 					double yDist = yPos - y;
 					double zDist = zPos - z;
--- a/src/main/java/rtg/world/gen/ChunkProviderRTG.java
+++ b/src/main/java/rtg/world/gen/ChunkProviderRTG.java
@@ -152,6 +152,7 @@ public class ChunkProviderRTG implements IChunkProvider
             for (int k = -parabolicSize; k <= parabolicSize; ++k)
             {
                 float f = 0.445f / MathHelper.sqrt_float((float)((j * 1) * (j * 1) + (k * 1) * (k * 1)) + 0.3F);
+                //float f = 0.445f / (float)Math.pow(((j * 1) * (j * 1) + (k * 1) * (k * 1)) + 0.3F,1.5);
                 parabolicField[j + parabolicSize + (k + parabolicSize) * parabolicArraySize] = f;
                 parabolicFieldTotal += f;
             }
@@ -324,6 +325,15 @@ public class ChunkProviderRTG implements IChunkProvider
 
     private static final int centerLocationIndex = 312;// this is x=8, y=8 with the calcs below
 
+    private boolean totalNotOne(float [] tested) {
+        float total = 0;
+        for (int i = 0; i < tested.length; i++) {
+            total += tested[i];
+        }
+        if (total<.999||total>1.001f) return true;
+        return false;
+    }
+
     public float[] getNewNoise(RTGBiomeProvider cmr, int x, int y, RealisticBiomeBase biomes[])
     {
     	int i, j, k, locationIndex, m, n, p;
@@ -332,7 +342,7 @@ public class ChunkProviderRTG implements IChunkProvider
     	{
     		for(j = -sampleSize; j < sampleSize + 5; j++)
     		{
-    			biomeData[(i + sampleSize) * sampleArraySize + (j + sampleSize)] = cmr.getBiomeDataAt(x + ((i * 8)-8), y + ((j * 8)-8)).biomeID;
+    			biomeData[(i + sampleSize) * sampleArraySize + (j + sampleSize)] = cmr.getBiomeDataAt(x + ((i * 8)), y + ((j * 8))).biomeID;
     		}
     	}
 
@@ -345,7 +355,9 @@ public class ChunkProviderRTG implements IChunkProvider
         		{
         			for(locationIndex = -parabolicSize; locationIndex <= parabolicSize; locationIndex++)
         			{
-        				hugeRender[(i * 2 + 2) * 9 + (j * 2 + 2)][biomeData[(i + k + sampleSize + 1) * sampleArraySize + (j + locationIndex + sampleSize + 1)]] += parabolicField[k + parabolicSize + (locationIndex + parabolicSize) * parabolicArraySize] / parabolicFieldTotal;
+        				hugeRender[(i * 2 + 2) * 9 + (j * 2 + 2)]
+                                [biomeData[(i + k + sampleSize + 1) * sampleArraySize + (j + locationIndex + sampleSize + 1)]]
+                                += parabolicField[k + parabolicSize + (locationIndex + parabolicSize) * parabolicArraySize] / parabolicFieldTotal;
         			}
         		}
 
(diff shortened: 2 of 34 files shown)


VelocityRa:
Awesome.
The optimization I was talking about can still be done in this code, I'll push it once Pink merges the PR.

VelocityRa:
Awesome.
One @whichonespink44 merges this I'll push the optimizations I was talking about.
To make things even faster, could we avoid checking for positivity in the `mix` functions?

Zeno410:
I'm thinking maybe we can forgo the complex staged mixing and just build the chunk biome % map directly from the neighborhood sampling.

VelocityRa:
Wait, isn't that how it's done already? I thought the mixing stuff was just for terrain smoothing, wasn't it?
(I'm in Discord if you wanna talk there)

Zeno410:
The smoothing is done from a biome map which has an "influence" from each biome for every chunk. It calculates the terrain from each biome that has "influence" on each chunk, then does a weighted average. That's the smoothing procedure.

All the mixing is to construct that block-by-block map. It starts with an every-8-block grid of the biomes in the neighborhood. Then it constructs an influence map for an every-8-block grid using a weighting function which is called "parabolic" but which is actually pretty linear (blocks twice as far away have half as much influence; there's an adjustment to prevent div0 at the actual block). Then it does mixes to make an every - 4 block grid, then an every 2, then an every block grid, then it uses the center of that for the chunk.

Zeno410:
Argh, as soon as I post the PR the deco changes start giving problems. Somebody may want to pull out ChunkProviderRTG and post it separately.

whichonespink44:
@Zeno410 Do you just want to change the decos back and re-submit the PR? If not, maybe @VelocityRa can pull out ChunkProviderRTG and submit that along with his other optmizations?

Zeno410:
I'd much rather he pulled it out than to try to revert all this stuff.

whichonespink44:
@VelocityRa You cool with that mate?

VelocityRa:
Sure, what will I need to do exactly? Pull this PR on my branch, make the changes and then submit a new PR?

whichonespink44:
If you can do that, I think that's probably the way to go, as it's only b3d8bff that we don't want from that PR I presume, correct?

VelocityRa:
@Zeno410 Waiting for confirmation on whether I should do that.
Also I did some profiling and unfortunately the function in which the optimization is (`mix4`), in runs for less than half a second (when generating a new world), so the performance boost is negligible... 
Also, Zeno was right, `populateChunks` (decoration) takes like 2x the CPU time the actual generation takes.

whichonespink44:
> (decoration) takes like 2x the CPU time the actual generation takes.

The only two things I can think of that would be causing a bottleneck during decoration are the increased number of Forge event postings per chunk (if there are 3 different trees/shrub decos in a biome, the TREE event gets posted 3 times in the new system, as opposed to only once in the old system) and the increased number of noise calculations in the new system. In the old system, one noise calculation would get called [at the beginning of the legacy rDecorate() method](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/master/src/main/java/rtg/world/biome/realistic/vanilla/RealisticBiomeVanillaForest.java#L51), but the new system calculates noise on a per-deco basis (assuming the deco utilizes noise), so if there are 5 different noise-based deco objects in a biome, the noise gets calculated 5 times per chunk (and often it's the same calculation).

Not sure how much of a performance impact either of those two things has, but that's pretty much the only thing that's changed in the new system that I can think of.

VelocityRa:
Take a look at this:
pic
`decorateInAnOrderlyFashion` is responsible for the majority of time `populateChunk` runs (only ~260ms is the rest here), so the Forge event postings apparently don't take that much time.
I can't analyze this much, you're the deco expert :P But I figured it might be a little bit useful.
There's a few Decos here but the one I've expanded, `DecoBaseBiomeDecorations` is the most prominent. (which makes sense I would imagine).
Also Minecraft's `BiomeGenHills.decorate()` is called, is this normal?

whichonespink44:
That looks like a Jungle biome, probably the worst culprit of them all. The `BiomeGenHills` is part of the base biome's decorations, so in this particular case, it might be worth re-decorating the Jungle biomes to avoid the `DecoBaseBiomeDecorations` calls.

Also, how is rPopulatePreDecorate() getting called *after* decoration?!

Hmm, I know it looks bad, and it definitely is, but... I suspect decoration's always been a resource hog. The question we need to answer is... is the new system worse than the old system? Cuz if it's not, then these new performance issues are being caused by something else.

VelocityRa:
Yeah I have no idea if it's faster than the old system.

Also, preDecorate isn't called after decoration, that's just how the profiler sorts functions (the ones that take longer come first). It takes a very short amount of time for that to execute so it's in the bottom.

Btw I love how this has sprung a discussion completely unrelated to the PR. :P

topisani:
You should try testing this and the old version (pre 0.8?) With the same seed and setup.

topisani:
Also what happens if you use lonely biome and try with two biomes with a lot of decorations where one is fully RTG decorated and one is only baseBiomeDecoration. Maybe it turns out we gain speed by using our system, and maybe it turns out our system sucks

VelocityRa:
Let's continue this discussion in #603.

whichonespink44:
@VelocityRa I think I'm gonna merge this PR, and then manually replace b3d8bff with the old stuff, and then push those changes to dev. Then you can just pull the latest dev at that point and PR whatever stuff you want.

In the meantime, just one more question for @Zeno410 - the EB GenLayer suppression in 4cac4dd, does this mean that EB now requires CC or BT?

Zeno410:
Um - well - even with the suppression the EB layers show up sometimes! But theoretically, it should.

src/main/java/rtg/world/gen/ChunkProviderRTG.java
```diff
@@ -874,16 +970,18 @@ public void populate(IChunkProvider ichunkprovider, int chunkX, int chunkZ)
...
     		}
         }
 
+        TimeTracker.manager.stop("Post-decorations");
+        TimeTracker.manager.start("Entities");
         if (TerrainGen.populate(this, worldObj, rand, chunkX, chunkZ, flag, PopulateChunkEvent.Populate.EventType.ANIMALS))
         {
-            SpawnerAnimals.performWorldGenSpawning(this.worldObj, worldObj.getBiomeGenForCoords(worldX + 16, worldZ + 16), worldX + 8, worldZ + 8, 16, 16, this.rand);
+            SpawnerAnimals.performWorldGenSpawning(this.worldObj, worldObj.getBiomeGenForCoords(worldX + 16, worldZ + 16), worldX, worldZ, 16, 16, this.rand);
         }
```
VelocityRa: `TimeTracker.manager.stop("Entities");` should be here, not below.
(maybe add track Ice generation as well?)

src/main/java/rtg/util/VoronoiCellOctave.java
```diff
@@ -130,7 +136,7 @@ public float noise(double x, double z, double frequency)
 			{
 
 				double xPos = xCur + valueNoise2D(xCur, zCur, seed);
```
VelocityRa: Should this be:
double xPos = xCur + valueNoise2D(xCur, **x**Cur , seed); ?

whichonespink44:
Ok, so I was testing this for a while last night (and by testing I mean generating worlds and flying around at high speed for 5 minutes or so in each one).

My overall experience is that things definitely *feel* faster. I didn't experience any stuttering, except for one time on world load, but other than that, everything just feels faster.

However... there's a massive amount of the following server warning [in the log](https://gist.githubusercontent.com/whichonespink44/f3d06645568d3eb804c1c09182f92379/raw/16175822009459381e9c363f1eaa96dd536a693e/latest.log):

```
[01:11:57] [Server thread/WARN]: * There is an attempt to load a chunk (-9,-2) in dimension 0 that is already being loaded. This will cause weird chunk breakages.
[01:11:57] [Server thread/WARN]: *  at net.minecraft.world.gen.ChunkProviderServer.originalLoadChunk(ChunkProviderServer.java:154)
[01:11:57] [Server thread/WARN]: *  at net.minecraft.world.gen.ChunkProviderServer.loadChunk(ChunkProviderServer.java:131)
[01:11:57] [Server thread/WARN]: *  at net.minecraft.world.gen.ChunkProviderServer.func_73158_c(ChunkProviderServer.java:101)
[01:11:57] [Server thread/WARN]: *  at net.minecraft.world.gen.ChunkProviderServer.func_73154_d(ChunkProviderServer.java:199)
[01:11:57] [Server thread/WARN]: *  at net.minecraft.world.World.func_72964_e(World.java:419)
[01:11:57] [Server thread/WARN]: *  at net.minecraft.world.World.func_147439_a(World.java:345)...
```

Other than that though, I don't know what voodoo you're doing Zeno, but it seems to be working well with just RTG on its. Just need to get to the bottom of the server warnings, and then if everyone's happy with this, I'll merge it to dev and we can release it into the wild to put it through its paces.

Zeno410:
Well, I found a different problem - ChunkProviderServer.chunkExists() doesn't actually do what it promises; it should be called chunkLoaded(). This causes less trouble than I would have expected - occasionally it skips decorating a few chunks, which is largely undetectable except in snowy areas. I was rewriting it to refer to the AnvilChunkLoader before I saw your message. Is there any more of the log? There's not enough there to see where it's happening in my code.

whichonespink44:
Full log is here: (it's massive)

https://gist.githubusercontent.com/whichonespink44/f3d06645568d3eb804c1c09182f92379/raw/16175822009459381e9c363f1eaa96dd536a693e/latest.log

I'm glad you mentioned the decoration skipping... I noticed it as well, but thought it was related to when I upgraded the decorations to the Deco system.

Zeno410:
It's a doubly bogus error. It's not from a chunk being *loaded* twice, it's from a chunk being *created* twice, so it's misnamed. And, of course, I already knew I could have recursive calls to create and just save the in-process chunk to return, so actually it all works out fine.I think I can do a workaround using ChunkLoad events.

Zeno410:
Incidentally, i suspect with this method we could further improve performance by rewriting the external biome decorators with the decorations centered on the chunk. Right now external decorators probably have the vanilla off-centering which creates a much higher chance of an out-of-ring decoration which can descend into the rabbit hole - especially with big trees, which almost guarantee it.

whichonespink44:
By 'external decorators', are you referring to when we call the base biome's decorate() method via DecoBaseBiomeDecorations? If so, then definitely think we should do that at some point. @VelocityRa was doing some profiling last night, and those base biome decorations are definitely a bottleneck.

whichonespink44:
That's gonna be a massive project though :cry:

Zeno410:
Well, it's a performance enhancement, not an obligation. Also, I think we probably only need to do if for biomes with large decorations, which is to say large trees. Things like WorldGenGrass can place right up to the outer rim boundary, which would set off the next chunk via neighbor updates, but that's only about 1 block of grass in a 1000, which is too rare to cause significant trouble. I pretty much only see trouble once I hit a large trees biome, although it seems like once I've hit one, the problems can stick around.

Zeno410:
*If* I'm right, remember, we will substantially *surpass* the performance of these mod biomes in Default, and that's a heck of a selling point. I certainly see some painful slowdowns in Highlands Default.

Zeno410:
Ok, fixes for the spurious error messages and the missed decorations. I flew around most of a maxed-out  map of an all-snowy world with no stutter, no missed decorations, and no generation pauses even through heavily-decorated Cold Taiga and the big trees Highland biome Tall Pines *until* I hit a large flatland Tall Pines and that still pretty much locked me up for a minute or two. Afterwards it recovered. So it seems to handle everything except large dense big trees biomes (mostly Tall Pines is mountainous and so the Firs are less dense and the technique can handle it). It would definitely be worth testing whether centering the Tall Pines (if they are offcenter) and seeing whether that improves performance.

This also has the candidate reversion to Voronoi noise for the rivers. IMO the rivers aren't as nice with Voronoi as they are with CellularSimplex but it should fix all those. obnoxious chunk wall/excess flatland/double river issues.

src/main/java/rtg/world/gen/ChunkProviderRTG.java
Zeno410: This is all part of centering decorations on the chunk. Everything is offset 8 blocks. SpawnerAnimals adds an additional random number onto the passed x and z, which is almost certainly 16. So the modified code means animals always spawn in the generating chunk.

VelocityRa: Oh, yeah, I know, I just said that the TimeTracker for entities should `stop` just below this line, and not below the Ice gen (maybe Github didn't show it for you, you have to press the button on the left in the diff page or just use the IDE). In other words, the tracker for Entities also counts the time spent on Ice gen and that shouldn't happen obviously.

whichonespink44:
Just testing these latest changes now... if everything looks good, I'll merge to dev.

whichonespink44:
Just pre-generated a world with ACT and noticed a different kind of console spam this time:

https://gist.githubusercontent.com/whichonespink44/dab6bca2c0914723d6db59c9ff800a2e/raw/1f116510f2661e636af739cb1491d423d1474b88/latest.log-entities

Here's some background info on those "Silently catching entity tracking error." warnings: https://github.com/MinecraftForge/MinecraftForge/issues/1648

I can confirm this is also happening when flying through the world normally:

https://gist.githubusercontent.com/whichonespink44/2b5f10fd51d6443f5b3742408c75bc61/raw/307f3904f0f5ad3a34ff3c523ac74ab23a389d86/latest.log-entities-gameplay

I don't know how big of an issue this is in the grand scheme of things, but it feels like this could potentially cause conflicts with mob mods, so probably worth getting to the bottom of it.

Also, it seems there are quite a few visible chunk boundaries in the terrain:







Could this be related to the Voronoi noise switch?

Here's the world map for reference:



World seed = 3042262550436252977

Zeno410:
The tiny chunk walls showed up when I got rid of the symmetrical dots. The symmetrical dots might have actually been that chunk walls scattered around. I suspect a weakness in the smoothing algorithm.

Entity problems, bleah, I don't know that system at all.

whichonespink44:
Potential 'gotcha' for reference:

> Apparently the entity tracker checks if an entity is already tracked by its `id`, not its `UUID`.

whichonespink44:
More entity-related braindumping...

Are we creating a 'fake' world for these buffer chunks? Cuz if so, Forge has an EntityJoinWorldEvent which is cancellable... maybe we could use that when we're creating the buffer chunks and cancel the event if the world is our fake world?

There's also an EnteringChunk event, but that one's not cancellable.

Zeno410:
No, the "fake" world is Topi's idea (although I'd thought about the same thing for a while). I'll take a look to see if that event could be used to stop the re-registration.

whichonespink44:
Ok, just giving this a test drive now.

PR #673, opened 2016-05-02
Entity reregistration spam fixed

Zeno410:


whichonespink44:
Nice work mate - it's well and truly fixed now... not a single entity warning, not even on startup :+1:

PR #674, opened 2016-05-02
New tree system.

whichonespink44:
Refactored all of RTG's custom trees, merging similar trees where possible and removing duplicate or unfinished legacy trees.

All RTG trees can be configured to use any log block and leaves block, including blocks from other mods. This means that we can generate a tall 'pine tree' made from Birch wood and Jungle leaves if we wanted to. Also, most trees can be configured to have different trunk heights and crown heights, and some even have additional parameters like the number of branches, etc. However, there are one or two trees that don't have any configuration options (except log block and leaves block) due to the way they have been coded. Also, all RTG trees have 'leafless' versions - just set the `noLeaves` option to `true` when configuring the DecoTree object.

Trees have been renamed according to their scientific 'binomial name', and future RTG trees must conform to this standard. This will allow for a robust inheritance-based tree system where sub-species can use the same root, trunk, branch and crown logic as their parent species where it makes sense to do so. It also provides an objective standard against which we can make design decisions as the system develops and our collection of trees grows. I considered creating aliases for the trees based on the their 'common names' to make it easier to work with the trees in the code, but decided against it in the end because a majority of the trees' common names were just as ambiguous as their binomial names. To help developers identify trees more easily, verbose Javadocs have been added to the constructors of each tree which include the common name, relevant configuration options, as well as inline screenshots for easy reference.

Also added a new DecoCollection system which allows for the grouping of deco objects for re-use across different biomes, mainly between biome variants (e.g. Forest, Forest Hills, Forest M, etc.).


topisani:
those javadocs sound sexy!

PR #683, opened 2016-05-04
Straight lines candidate fix

Zeno410:
Also conversion of scenic lakes to river/frozen river, with config option to substitute biome.

This is "for discussion and review" as the fix changes the character of biome transitions, and it could impact performance (either way).


whichonespink44:
Ok, so I've been testing this PR, and it's looking really good. Didn't come across any straight line/chunk border things; however, I *have* discovered some weird terrain issues:


Seed = -985965778539080818
x = 2226, z = -1514


Seed = 1441114414487066402
x = 529, z = -62

Not sure what's going on there (surprise!), and it's not a show-stopper, but I suspect it's probably a sign of something else going on.

In addition, I keep finding undecorated chunks. @Zeno410 I know we discussed this, and I know that RTG has issues around biome borders, but... it's significantly more noticeable than it was before and sticks out like a sore thumb. Again, it's not a show-stopper, and if that's the compromise we have to make in order to have acceptable levels of performance, then I definitely choose the latter, but if there's anything we can do, I think it's definitely worth exploring. It's mainly the abrupt halting of grass gen that makes it look so bad. Is there any way of making the grass go just a little bit further without slipping down the rabbit hole?

Also, I've tested the new scenic lake biome config options and they're working fine as far as I can tell. That's going to make a lot of people happy for sure.

All in all, it's working great - just a few niggles to sort out.

Zeno410:
I'm pretty confident the grass business isn't undecorated chunks. I think I rewrote WorldGenGrassRTG or whatever it's call to stay entirely in its chunk earlier (I commented out some random jittering). That's probably not needed anymore. What may be happening - check the code - is the the grass isn't always getting placed, and whether it's placed is on a per-chunk basis.

whichonespink44:
Ok cool, in that case, I think I'll just leave DecoGrass as it is until we get these other issues stabilized (don't want to do something wrong in DecoGrass and add confusion to the debugging process).

Zeno410:
I've found and fixed the problem of undecorated chunks but it seems I can't push *that* change until *this* one gets merged.

whichonespink44:
Merged in cf2e13e.

I updated my local branch of this PR from dev and made a minor change to the scenic lake config, so thought it would be best to merge that branch. Everything's in dev now, so feel free to PR the undecorated chunk fix and I'll get that into dev as well.

PR #692, opened 2016-05-06
Minor update

Zeno410:


whichonespink44:
@Zeno410 Do you think it's worth me going through all the RB files and getting rid of the static variables for the top & filler blocks and moving them into the constructor like you did here? https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/dev/src/main/java/rtg/world/biome/realistic/highlands/RealisticBiomeHLSteppe.java#L21

Also, why does BiomeBase set the top & filler blocks? Just a sanity check?

Zeno410:
The BiomeBases are BiomeGenBases and there's an implied contract that a biomeGenBase have top and filler blocks. So I figure it should have it. In addition, if anything really uses all those topBlocks then they'll be there.

I think the copies of the top and filler blocks should be gone in principle, but I doubt it's a big issue. I figure I should take them out as long as I'm doing other changes.

whichonespink44:
Cool, I'll do the same then... just take em out as I come across them when re-decorating or whatever.

PR #693, opened 2016-05-07
Smoothing changes and speed improvements

Zeno410:
Made the smoothing weight nearby biomes more and removed 256 float[256]
allocations from each chunk generation.

### DIFF
--- a/src/main/java/rtg/util/Converter.java
+++ b/src/main/java/rtg/util/Converter.java
@@ -1,0 +1,81 @@
+
+package rtg.util;
+import java.util.*;
+
+/**
+ * A class for converting one type into another
+ * @author Zeno410
+ */
+public abstract class Converter<FromType,ToType> {
+    
+    /** Creates a new instance of Converter */
+    public Converter() {
+    }
+    
+    abstract public ToType result(FromType original);
+    
+    public ToType of(FromType original) {return result(original);}
+    
+
+    public Collection<ToType> listResult(Collection<FromType> source) {
+        Collection<ToType> result = new ArrayList<ToType>(source.size());
+        Iterator<? extends FromType> sourceList = source.iterator();
+        while(sourceList.hasNext()) result.add(result(sourceList.next()));
+        return result;
+    }
+    
+    class IteratorConverter implements Iterator<ToType> {
+        Iterator<? extends FromType> source;
+        IteratorConverter(Iterator<? extends FromType> theSource) {source = theSource;}
+        public boolean hasNext() {return source.hasNext();}
+        public ToType next() {return result(source.next());}
+        public void remove() {source.remove();}
+    }
+
+    
+    public Acceptor<FromType> acceptor(final Acceptor<ToType> target) {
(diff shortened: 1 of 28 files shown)


Zeno410:
This changes the smoothing, with sometimes noticeable changes in terrain. Canyons and mesa pop up much closer to borders. Small Hills areas can be much taller. Other team members should look at how this changes things.

If we want to go with this, I'd like to reduce the heights of hills sub-biome to cut back on the "parabolic" effect. In the past I saw problems with this and Canyons connecting to mountains but I just checked a few and they were acceptable.

whichonespink44:
Just been doing some testing with this PR and could definitely see the difference in Mesas and Savanna Plateau where the elevated terrain appear much closer to the biome border than before, which looks fine.

Also noticed that hills sub-biomes tend to become 'Hills' much higher up the hill than they used to. I'm guessing that lowering the hills will help with this?

All in all, I couldn't see that much of a difference, so if you think this is the way to go, then +1 from me.

Just gotta ask though... what's the main benefit(s) of changing the smoothing? What problem are we trying to solve?

Zeno410:
The main problem is Topi's complaint that the hills are too "paraboloid" with the previous new smoothing. This smoothing makes it better in some ways and worse in others, but if we step down the base heights of the Hills sub-biomes it should be mostly a win.

In addition it reduces the "blah effect" at biome borders. There's less intermediate territory which has a lot of influence from multiple biomes, which tends to become a constant slope.

A last benefit is that terrain sub-biomes now generally have noticeable effects where before they were more suppressed (and the smoothing system in dev now is already better than the legacy one).

If you're good to go with this I'll go through and push down the base heights of the hills sub-biomes. I probably won't have those in until late night my time though.

whichonespink44:
Ok, I think that all makes sense - well if you guys are happy with it, then let's roll with it.

No worries about getting it in late - I'll be in 1.9 mode today/tomorrow so no rush.

whichonespink44:
@Zeno410 Is this ready to be merged? If so, I'll get it into dev later on today.

Zeno410:
Can somebody else test the speed for me? It seems slower after some changes trying for Streams compatibility even though they shouldn't make a difference. Tp's in particular are working oddly (long pauses before generation).

whichonespink44:
After flying and TP'ing around for about half an hour, I didn't notice any significant difference in speed compared to the current dev, including TP's (which, to be fair, are always a bit dodgy on my laptop), but they weren't any more dodgy than they normally are.

The only times it stuttered at all was once or twice as I approached a biome with a lot of trees (Jungle & Roofed Forest), and a couple times after TP'ing.... but in general it seemed fine to me.

I only tested with RTG on its own though. Did you want some feedback when using Streams or another biome-adding mod?

Zeno410:
It doesn't work with Streams at present - I don't need to know that. Other biome mods should be fine. I'd like to know what happens with generation mods but we need a release candidate for that - there's too much for us to test. I do suspect a problem with Mariculture from a forum post, but it's hard to be sure.

PR #702, opened 2016-05-13
Streams incomplete decoration fix

Zeno410:
At least on my machine. Might also fix the Witchery villages #686.


whichonespink44:
Ignore the conflicts - I'm on it.

whichonespink44:
Created new branch from this PR and pulled from dev:

https://github.com/Team-RTG/Realistic-Terrain-Generation/tree/pr/702

Testing with Streams now.

Zeno410:
Speed improvements added and they're fairly major.

whichonespink44:
Reduced chunk gen time by two-thirds? Nice one!!

Just testing now.

whichonespink44:
@Zeno410 Preliminary testing is showing really good results so far. Performance has noticeably improved, stuttering only in Jungle biomes, even after flying around high-speed for 20 minutes. Even non-jungle biomes with lots of trees aren't stuttering for me.

Is there anything else specifically I should be looking out for?

whichonespink44:
2016-05-14_12 58 27_new world2_overworld_day

Streams look a little more squiggly than I remember, but... on the whole, it looks pretty good. And it was the first map I've been able to pre-generate since I don't even remember, so performance has definitely improved significantly. Before, I would get to about 10,000 chunks before MC froze.

Zeno410:
The freezes were probably due to memory issues; my chunk cache wasn't working properly, and I fixed that in the latest versions. Streams (for me) substantially slows down generation, BTW. At this point there's not much point in speeding up the RTG terrain because it's back to a few percent of total time, and only about 1/5 of deco/populate time. This might be the first time that's really true, because in the old estimates I ran the decoration time also included terrain generation that was set off by going down the rabbit hole. The current estimates are entirely honest - such rabbit holing as happens takes place elsewhere and doesn't affect these time estimates.

Speeding up decorations in vanilla, if possible, would require some advanced hackery. I think most deco time is spent in the elaborate notifications of adjacent blocks. We could short-circuit that by hacking into the chunk data, but it's not fun. Modded biomes are still using +8, which induces a lot of rabbit-holing in big tree biomes, so we could speed them up by RTG-ing the decor. 

A lot of time is spent on animal generation - we could suppress some of that since it seems something I did has an effect of making more animals.

whichonespink44:
> Speeding up decorations in vanilla, if possible, would require some advanced hackery. I think most deco time is spent in the elaborate notifications of adjacent blocks. We could short-circuit that by hacking into the chunk data, but it's not fun.

Couldn't we also RTG the decorations in vanilla biomes to get around the rabbit holing? It would mean quite a lot of manual decoration, but it might be better than hacking the chunks and potentially introducing new issues.

All the other performance-enhancing suggestions sound good, though I still don't see what you could've done to make animals generate more. ¯\_(ツ)_/¯

As for this PR, is it ready to be merged into dev?

Zeno410:
For the life of me I can't see a reason the animals would have increased, unless there's a bug in vanilla that causes a lot of animals to get "lost" in generation (which might have never been noticed, because there are still plenty). But it sure feels like they have.

I thought you already RTG'd most of the vanilla decor? But I was referring to the way vanilla does decorations - every block sets off a full set of check on the neighbors, which is unnecessary at least 99.9% of the time, to the point where if you just didn't do it things should be acceptable. With all the ore gen and the big trees that's a lot of work.At this point, though, our decor should be taking the same time as vanilla in Forge except for the size of the trees, so I don't see improvements as required.

If you don't see anything outrageously wrong, I think this is good to merge.

whichonespink44:
Everything from this PR (except for a31986a, which I missed, sorry) has been merged to dev in https://github.com/Team-RTG/Realistic-Terrain-Generation/pull/707

On a sidenote, in the future, I think we should probably start using remote feature branches instead of adding new changes to PR's. The last couple of PR's I had to merge into dev manually because they weren't in sync with dev, and really, we should probably only ever add minor tweaks/bug fixes to PR's. It's just gets a bit too 'scope creepy' (except instead of scope creep, it's more like 'PR change creep')

Since everyone has write access to the repos, anyone can create a remote feature branch from dev, and push stuff to those, pulling from dev when relevant, and then we can all reference and make comments on those branches instead of the PR's, if that makes sense?

Sorry, re-reading that, I probably haven't explained myself very well, but... essentially, let's try to not add new stuff to PR's unless it's a bug fix or a minor tweak. :yum:

PR #708, opened 2016-05-16
More Volcano fixes/tweaks

VelocityRa:


whichonespink44:
Nice one - I'll give this a test drive now :car: 

Also, what do you think about increasing the rarity of volcanoes? I think 6 probably makes them too frequent, but haven't been able to decide on what we should change it to.

VelocityRa:
I was gonna ask about that too actually, I figured *you* might have an idea since you've been testing them way more than me (I basically only see one that gets regenerated, so I don't know their current rarity).
Should they be 2 times rarer, 3?

whichonespink44:
yeah, I think 3 times as rare is probably a good place to start... I haven't actually messed around with the rarity in the configs yet, but I'll do that when I'm testing this PR

Maybe something else to look at at some point is... maybe some sort of limit on the number of volcanoes that appear in the same biome... cuz even by making them 3 times as rare, if you stumble onto a massive Jungle biome, you could still probably find 4-6 volcanoes there, some of them right next to each other, which looks kinda weird, but not a show-stopper.

whichonespink44:
Btw, don't forget to pull from dev at some point... I just checked out this PR and pulled from dev to make sure it's using the latest stuff. Don't worry about littering the commit history with branch merges, etc. I actually like seeing those messages :)

VelocityRa:
Ah ok!

VelocityRa:
Not sure what happened there (the extra commits), but I'm ahead of upstream/dev now

whichonespink44:
@topisani oh wise and powerful Git Guru of the interwebz... any idea why Velo's getting duplicate commits when pulling from dev?

VelocityRa:
I'll figure it out eventually :P I think I know what I did wrong.

whichonespink44:
So I've been doing some testing, and volcanoes are looking great. I was a bit skeptical about the new lava/eruptions, but having seen them in the wild, I'm now fully converted. :+1: 

Also, I think I found the magic defaults for volcanoe chance (32 for Mega Taiga variants, 64 for Jungle variants). Might be a little too rare, but when they do generate, they should almost always be a one-off volcano.

I've pushed those volcano config tweaks to dev... I think the only things left now are the mix blocks and blending in the volcano base with the host biome's surface (if possible).

etc/config/RTG/rtg.cfg
```diff
@@ -330,8 +330,8 @@ volcanoes {
 
     # 1/x chance that a volcano will generate in a biome that has volcanoes enabled.
     # 1 = Always generate if possible; 2 = 50% chance; 4 = 25% chance
-    #  [range: 1 ~ 2147483647, default: 6]
-    I:"Volcano Chance"=6
```
VelocityRa: The description here and in some other places is wrong, we ought to regenerate the configs.

VelocityRa:
If you don't have any more tweaks coming shortly, we should probably merge this, these commits re-appearing are getting crazy and I noticed some unrelated ones here too.
The mix block surfacing stuff can come in another PR.

PR #728, opened 2016-05-26
Stargates from SGCraft now generate in desert temples.

whichonespink44:


Ezoteric:
pinky you tha man as usual :P

PR #743, opened 2016-05-31
Merge dev into master (0.8.1).

whichonespink44:


whichonespink44:
@Zeno410 Before I release 0.8.1, can you confirm if updating from 0.8.0 to 0.8.1 will cause chunk walls? I had a look through all the commits and didn't see any changes to terrain, but I don't know if anything from CPRTG might affect this.

whichonespink44:
I've tested several worlds for chunk wall issues and found a grand total of zero, so it looks like this one's good-to-go :+1:

srs-bsns:
Was a fix for the ocean floor issue in one of these commits?

whichonespink44:
> Was a fix for the ocean floor issue in one of these commits?

No, I think we're saving that one for 1.0 as it could create minor chunk wall issues.

PR #749, opened 2016-06-12
Papering over ocean floor bug

Zeno410:
Apparently caused by a small rounding error in the biome weighting. The
effect is less that 0.01, so completely unnoticeable apart from flat
areas right at a block cutoff. Possibly an unavoidable error due to
floating-point precision limits.

### DIFF
--- a/src/main/java/rtg/world/gen/terrain/TerrainBase.java
+++ b/src/main/java/rtg/world/gen/terrain/TerrainBase.java
@@ -8,7 +8,7 @@ public class TerrainBase
 {
     protected float base; // added as most terrains have this;
     protected final float minOceanFloor; // The lowest Y coord an ocean floor is allowed to be.
-    public static final float minimumOceanFloor = 30f; // The lowest Y coord an ocean floor is allowed to be.
+    public static final float minimumOceanFloor = 30.01f; // The lowest Y coord an ocean floor is allowed to be.
     protected final float minDuneHeight; // The strength factor to which the dune height config option is added.
     public static final float minimumDuneHeight = 21f; // The strength factor to which the dune height config option is added.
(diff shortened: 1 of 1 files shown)


srs-bsns:
Tested and confirmed fixed.

Before:

After:

whichonespink44:
Nice work fellas!

I'm going to see if this change causes chunk walls if updating an existing world.

```
Seed: 5955455629908755995
x: 200, z: -500
```

whichonespink44:
Tested two different 0.8.1 worlds and didn't see any chunk walls. I did see one instance of some artifacty/jittery weirdness when crossing between existing->new chunks, but... it was barely noticeable and only spanned two chunks.

So far, it looks like 1.0 will be backwards-compatible with 0.8 worlds. :+1:

Zeno410:
Where did those occur? Particular biomes, at biome transitions, ?

whichonespink44:
RTG + BOP (2.1.0.1826) + BT (default example tweaks)

Seed: 6885086430968145248

Pregenned a world with ACT (pregenspawn 100) using RTG 0.8.1 and then replaced it with the latest dev to check for chunk walls, and this was the only slightly buggy thing that I found along the genned/ungenned edges. It's a BOP Orchard biome along the edge of the mouth of an ocean to the north, but... after looking into it further, this isn't being caused by updating from 0.8.1 to the latest dev because I just created a completely new world with the latest dev and flew to those coords, and it looks exactly the same.

Zeno410:
That looks like the river edge bug that was reported a while back. I don't think I fixed it, actually. Do we have an issue for it?

whichonespink44:
Nope, must've slipped through the cracks. Just testing the latest dev now with your new PR so hopefully that will sort it.

PR #750, opened 2016-06-12
Volcano shelves over water fix

Zeno410:


Ezoteric:
screenshot or it didnt happen ;)

PR #754, opened 2016-06-13
Vanilla saplings grow RTG trees.

whichonespink44:
oak_sapling

As a special 1.0 bonus feature, I've been thinking about adding the ability for vanilla saplings to grow into RTG trees (what was previously discussed as the 'RTG Saplings' addon).

I've already got a very basic proof-of-concept to demonstrate the approach, and it even works in some biomes (kinda). If you bonemeal any sapling, it sometimes generates an RTG tree that's been registered in that biome's `decos` list (confirmed working in Extreme Hills and Birch Forests). The only catch (when it 'works') is that you have to log out and log back in again to actually see the tree. So obviously there's a lot more work to be done, but it looks like it's possible.

I've created this PR so you guys can test it out if you want, and also to open up a discussion as to whether this could live in the core RTG mod, or whether it would be better suited as an addon mod.


Zeno410:
Hunh, I thought this would be pretty hard.

Ezoteric:
testing now, and if it works its cool, not all trees grow the same in rl also so this would be a nice addition where the addon can always make the custom trees with custom sapplings.

so for the last time asking, give me a base mod so i can work on this addon :P

Ezoteric:
forest aint working, neither does jungle so far.

jungle is pretty weird>

Ezoteric:
roofed forest works with the dark sappling 4x

taiga works but had to relog and the sappling stays

whichonespink44:
Only Birch Forest and Extreme Hills were confirmed working in the current version, but I have some new changes I'll be pushing tonight where all vanilla biomes will be working properly.

Ezoteric:
didnt know you checked all biomes :P

whichonespink44:
As of the last batch of changes, a sapling will generate a random RTG tree from the list of trees that are associated with the biome in which the sapling has been placed. All RTG trees have been setup, so this should work in any biome that has RTG trees associated with it.

Currently, all saplings produce this effect, but I'm going to make it so that each tree has a particular sapling associated with it, and that tree will only generate if the correct sapling is used. For example, in the Forest biome, the darkwood tall tree might require a spruce sapling, and the lightwood tall tree might require an oak sapling, etc.

Also, I've found a bug that I'm not sure how to fix... sometimes the biome that I'm getting from the event coords isn't the biome that it says I'm in. For example, if I'm in a Forest biome, when the tree generates from the sapling, sometimes it says 'River' instead of Forest (even though the biome in F3 says 'Forest'). I'm assuming this has something to do with biome repair? If so, does anyone know how I can get the actual biome from the block X/Y/Z coords and not the biome that has been repaired?

Zeno410:
I think the chunk can give you the final biome. Sounds like something is evading the river correction.

whichonespink44:
Ok, so I think this is just about finished.

I'm currently going through and testing all of the biomes where RTG's custom trees generate naturally to see if the saplings work. Assuming that goes well, the only issue to resolve is the wrong-biome bug...

@Zeno410 This is how I'm currently getting the biome...

```
WorldChunkManagerRTG cmr = (WorldChunkManagerRTG) world.getWorldChunkManager();
BiomeGenBase bgg = cmr.getBiomeGenAt(x, z);
RealisticBiomeBase rb = RealisticBiomeBase.getBiome(bgg.biomeID);
```

... where X and Z are the block-X and block-Z coords provided by the SaplingGrowTreeEvent event.

Is there another way to get the biome from those block coords?

Ezoteric:
PMed you the results,

btw do we also have custom trees in modded biomes ?

whichonespink44:
Cool, I'll double-check the ones you mentioned in your testing results.

And yes, there are custom trees in modded biomes as well. Highlands has a bunch, EB has a few, BOP has a couple... even AbyssalCraft has some. Easiest way to check is to do a search for 'DecoTree'.

whichonespink44:
@Team-RTG/everyone Now that this feature is complete (except for a couple of minor issues), how does everyone feel about adding this to the core mod, rather than making it an addon?

Also, is there anything about the functionality that we should change? Here's how it works at the moment:

**Config options**
```
saplings {
    # Set this to TRUE to allow RTG's custom trees to generate from vanilla saplings.
    # RTG's custom trees require a sapling that their leaves drop.
    # RTG's custom trees can only be generated in biomes where they naturally generate.
    #  [default: true]
    B:"Enable RTG Saplings"=true

    # 1/x chance that a vanilla sapling will generate one of RTG's custom trees.
    # 1 = Always generate if possible; 2 = 50% chance; 4 = 25% chance
    #  [range: 1 ~ 2147483647, default: 2]
    I:"RTG Tree from Vanilla Sapling Chance"=2
}
```
**Basic functionality**
So basically, RTG's custom trees can be grown only from the saplings their leaves would drop naturally, and only in the biomes where they generate naturally. For example, you can't grow a Swamp Willow in anything except a Swamp biome (and only with an Oak sapling, because it has Oak leaves). They even adhere to their own generation rules, so... in the case of the Swamp Willow, the sapling has to be near water or it won't grow because Swamp Willows only generate naturally near water.

**Vanilla trees**
As for how this affects vanilla tree growth, the user has a few options with the `"RTG Tree from Vanilla Sapling Chance"` config option. By default, there's a 50/50 chance that a custom tree will **try** to grow. 50% of the time, it won't even try to grow a custom tree and just grow a vanilla tree. The other 50% of the time, it will grow a custom tree only if all of the conditions (right sapling, right biome) are correct. If the conditions aren't right, a vanilla tree will grow instead.

**Under the hood**
Most of the magic happens in EventManagerRTG, so it's pretty self-contained. The only refactoring that had to be done consisted of creating a new ArrayList to store the configured trees for every biome. This involved tracking down all instances of DecoTree and configuring the tree before creating the deco. Sounds more involved than it was, and even though a lot of files have changed, there wasn't any voodoo involved. Initially, I tried to just iterate through all of the Deco objects that have been added to the biome, but that ended up being a lot more complicated than I thought it would because of the DecoHelpers and DecoCollections. I tried to iterate through those as well to pull out just the tree decos, but... there were other complications there as well, and it started getting really messy, so I just decided to create separate lists for the trees. All other deco-related functionality is the same as it was before.

whichonespink44:
Ok, I've taken this about as far as I can, so if everyone's cool with this being part of core rather than an addon, then I'll merge it into dev later today.

The only bug remaining that I'm aware of is that sometimes the biome checking isn't accurate, usually when the sapling is near a body of water, so if anyone wants to lend a hand there, that would be groovy.

The only other issue that I suspect we might encounter is the way the `generateFlag` variable is being set in `onSaplingGrowTree`. It works fine in singleplayer, but I have a feeling we might run into trouble on servers when, for example, someone is bonemealing a sapling whilst another person is exploring new chunks. It feels like there might be some race conditions to deal with, but hopefully I'm wrong. Only way to know for sure is to release it out into the wild and see what happens.

whichonespink44:
```
this.saplingBlock = Blocks.sapling;
this.saplingMeta = (byte)0;
```

Sticky note for later... the above variables in TreeRTG are unnecessary now. Need to remove those before merging to dev.

UndeadZeratul:
Is there a plan to add this feature to the 1.9 version?  Just curious.

whichonespink44:
> Is there a plan to add this feature to the 1.9 version? Just curious.

Yes, we'll be porting the sapling functionality across to the 1.9.4 version, probably shortly after we sort out the performance issues.

UndeadZeratul:
Ah, cool.  Performance is more important, especially considering custom trees like this would do nothing but hurt it further.  Glad to hear this will eventually make it there, though!

PR #771, opened 2016-06-24
Village-aware decos.

whichonespink44:
This is the stuff I've done so far on the village-aware decos.

The village flag has been renamed to `hasPlacedVillageBlocks`, and it is now being passed to all decos. However, only trees, shrubs, logs, and boulders are utilizing it, since they're the most troublesome ones in villages.

I'm putting this up for review because @Zeno410 mentioned something about implementing a village cache of some sort and just want to make sure this stuff is going to be compatible with it, or if it's even necessary at all with the village cache.

Here are some before & after screenshots (sorry, the angles aren't exactly lined up):

village-1

village-2

village-3


Zeno410:
Very distinct improvement; looks really good!

whichonespink44:
Coll, I'll merge this later today after I make a few tweaks.

Issue #784, opened 2016-06-25
One repo to rule them all


whichonespink44:
@Team-RTG/everyone 

I propose that we consolidate the 1.9 repo into the Realistic-Terrain-Generation repo, as we've done with the 1.8.9 repo, and make a few minor changes to our branching/development workflow.

Essentially, here's what I'm thinking this repo should look like:

**Branches:**

```
1.7.10-master
1.7.10-dev
1.8.9-master
1.8.9-dev
1.9.4-master
1.9.4-dev
1.10-master
1.10-dev
Etc.
```

This branch structure would allow us to maintain our current master/dev workflow, whilst providing a central location for all of RTG's code.

The dev branches will continue to be used for pull requests and for staging the latest development versions, and the master branches will be frozen after every release for that version.

I know the directory/code structure between 'v1' and 'v2' is radically different, but it really doesn't matter to Git, and it would make my life as project manager easier if everything was in one place, and it would make it easier for users when submitting bug reports, etc, so... unless this setup is going to negatively affect anyone's local development workflow, or if there are any other good reasons to _not_ do it this way, then I think this is the way to go, especially with Mojang & Forge releasing more frequently in the months ahead.

One mod. One repo. Multiple version branches.

Thoughts?

topisani:
i am kinda torn on this.  
first of all that is in theory the way to go, but i prefer the idea of an archive repo for 1.7/1.8, simply because i think a refresh is nice, and this is one of the best chances we will ever have for that.

Also, with the talk we have had recently about having a very small core wrapping system that interacts with forge and minecraft version specific things, so we can use the same code across multiple game versions, except for one distinct package that is specific to a version/api (sponge/bukkit support could be considered this way), we might have to do some restructuring anyway. So i think we should take that discussion first.

Zeno410:
I have yet to get into the 2.0 code but my impression is that it's different enough it should be a different repo. The point of having things in the same repo is that git can automatically match up similar code and that's not going to work 1.0 -> 2.0.

whichonespink44:
> we might have to do some restructuring anyway

I'm up for restructuring the code to enable/improve compatibility with other mods/platforms, but I don't see why that would require different repos. Unless you meant a different kind of restructuring?

> The point of having things in the same repo is that git can automatically match up similar code and that's not going to work 1.0 -> 2.0

[...]

PR #785, opened 2016-06-26
Minestrappolation support

whichonespink44:
Putting this up for review to demonstrate the version branch workflow in case it wasn't obvious from #784. Also, whilst I'm pretty sure the 'one repo' approach is the way to go, the only way to really know for sure is to try it out and see if it falls down anywhere.

As for the PR itself, so far this only adds biome support for Minestrappolation (Redwood Forest & The Frost). The person who requested this [in the forums](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/2524489-realistic-terrain-generation-rtg-realistic-biomes?comment=2025) mentioned that a lot of world gen wasn't being generated, including MS ores, and I know why:

https://github.com/MinestrapTeam/Minestrappolation-4/blob/master/src/main/java/minestrapteam/mods/minestrappolation/world/MGenHandler.java#L120-L187

Our old friend, the biome comparison check (`biome == BiomeGenBase.xxxxx`)  :/

So whilst we could add biome support for Minestrappolation (either with the manual approach in the PR, or the welding approach from v2), we won't be able to fully support Minestrappolation until those biome comparisons checks are replaced with biome ID checks.

### DIFF
--- a/etc/config/RTG/biomes/minestrappolation.cfg
+++ b/etc/config/RTG/biomes/minestrappolation.cfg
@@ -1,0 +1,66 @@
+# Configuration file
+
+biome {
+
+    minestrappolation {
+
+        redwoodforest {
+            #  [default: true]
+            B:"Allow Villages"=true
+
+            #  [default: true]
+            B:"RTG Decoration: Logs"=true
+
+            #  [default: ]
+            S:"RTG Surface: Filler Block"=
+
+            #  [default: ]
+            S:"RTG Surface: Filler Block Meta"=
+
+            #  [default: ]
+            S:"RTG Surface: Mix Block"=
+
+            #  [default: ]
+            S:"RTG Surface: Mix Block Meta"=
+
+            #  [default: ]
+            S:"RTG Surface: Top Block"=
+
+            #  [default: ]
+            S:"RTG Surface: Top Block Meta"=
+
+            #  [default: true]
+            B:"Use RTG Decorations"=true
+
+            #  [default: true]
+            B:"Use RTG Surfaces"=true
+        }
+
+        thefrost {
+            #  [default: true]
+            B:"Allow Villages"=true
+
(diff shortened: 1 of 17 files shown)


topisani:
we could probably do the RBB != BGB thing in 1.8.9 too, it was fairly simple. But im not sure its worth the effort, does anyone play 1.8.9 anymore?

whichonespink44:
So I've been working on a PR for the Minestrappolation project, converting all of the biome checks to biome ID checks, but in doing so I noticed something else that is preventing their world gen from generating in RTG worlds:

https://github.com/MinestrapTeam/Minestrappolation-4/blob/master/src/main/java/minestrapteam/mods/minestrappolation/world/MGenHandler.java#L54

```
        if (chunkGenerator instanceof ChunkProviderGenerate)
        {
            this.generateOverworld(world, random, chunkX, chunkZ);
            if (Config.genBiomeStones == true)
            {
                this.genBiomeStone(world, chunkX, chunkZ, random);
            }
        }
```

So basically their entire world gen relies on the chunk provider being a subclass of ChunkProviderGenerate, which CPRTG is not.

So even if we implemented the BGB/welding system, most MS world gen would still be absent.

As for people still playing 1.8.9... Thaumcraft, which doesn't have a 1.9.x version yet, had over 100k 1.8.9 downloads in the last 3 months, so... there's at least that many people playing it. :stuck_out_tongue: 

As for RTG users, I don't know how many of our 40k 1.8.9 downloads are still active, but... all I know is I've promised several people that I would add support for Minestrappolation, so I'd like to keep that promise if possible (although it looks like it's going to be a lot trickier than I thought it was going to be). :disappointed:

whichonespink44:
Ok, maybe not as tricky as I thought! :dancer: https://github.com/MinestrapTeam/Minestrappolation-4/pull/121

PR #787, opened 2016-06-26
RBB != BGB

whichonespink44:
RealisticBiomeBase no longer extends BiomeBase, and therefore no longer extends BiomeGenBase.

I've removed BiomeBase completely, but we can always refer to it in Git's history if we ever need anything from it.

Let me know if you guys can spot anything I might have done wrong, especially in CPRTG, etc.


topisani:
:+1: looks good to me

Zeno410:
I'm not going to even bother to look. If this can run with a biome mod without going boom, it should be fine.

Issue #392, opened 2016-01-07
Desert villages use cobblestone structures


KeepOnDigging:
Villages generated in desert are made up of standard cobblestone structures:
![image of desert village](http://s5.postimg.org/sbmvpz00n/2016_01_08_09_30.jpg)

They should be made up of sandstone structures, as in Vanilla.

KeepOnDigging!

whichonespink44:
@KeepOnDigging - forgot to ask, is this happening with just RTG, or does it only happen when used in conjunction with biome-adding mods? If the latter, which one's the culprit?

Ezoteric:
I have only found a desert village made with cbbl when it was right next to a non desert biome and even then half the village was still in sandstone. I do not have cbbl villages in ny deserts in my world

whichonespink44:
Yeah, that would explain it actually... I believe the village algorithm works the same way as the structure algorithm in that it just needs to find a viable block in a viable biome, and then it can start generating the village up to 20(?) blocks away from that block. (I think)

KeepOnDigging:
I set up a world which generates only desert… and rivers.
This village generated with no other biomes for hundreds of blocks:
![image of desert village](http://s5.postimg.org/ej7387ufr/Screen_shot_2016_01_21_at_3_41_24_PM.jpg)

I don't have CC installed, or any other mods which add biomes.
RTG has:
    I:"Generate only this biome ID"=2
    B:"RTG Biome Layout"=true

just to make it generate only desert (a nice feature which I will miss when RTG layouts go away!).

Is there anything else I can do to help diagnose this problem?

KeepOnDigging!

KeepOnDigging:
I have just confirmed this bug with CC (latest) and snapshot5-rc1.

Ezoteric:
Is this with ONLY using rtg also ? Cause I may know why it's happening then and also why I don't have it :p

KeepOnDigging:
It happens with RTG only installed (i.e. using RTG to restructure and redecorate vanilla biomes). 

It also happens with CC installed, but no other biome-adding mods.

Ezoteric:
With only I mean it be the only mod installed besides forge :)

KeepOnDigging:
Yes.

With RTG as the ONLY mod installed, villages in deserts are made up of cobblestone buildings, never sandstone.

whichonespink44:
[...]

PR #820, opened 2016-07-09
New event handling system.

whichonespink44:
So I've been looking into RTG's event management issues, and I think this might be a decent approach to use, so I'm uploading this for review.

Essentially, instead of having all of the event handlers as methods in one class (EventManagerRTG), we need to move them out into their own classes (or as inner classes within EventManagerRTG). Doing this allows us to use the World.Load and World.Unload events to register/unregister the handlers, ensuring that RTG's event handlers are only registered in the RTG world type. The only event handlers that are ALWAYS registered are the World.Load and World.Unload handlers because they're the ones doing the registering and unregistering.

So the process looks something like this:
- preInit -> Register all event handlers.
- onWorldLoad -> Register all event handlers (if they're already registered, they don't get added twice)
- onWorldUnload -> Unregister all event handlers (except onWorldLoad and onWorldUnload)

This accounts for all the different scenarios where RTG's event handlers should/shouldn't be registered. The only place this approach might fall down is when dealing with multi-world servers, but I'm not exactly sure how that stuff works, so maybe there's a way around it, maybe there isn't.

In any case, there's still a bit of work to do with this PR, ~~for example, decorations aren't working properly (but I think I know why - just haven't gotten around to fixing it yet)~~, but I thought I'd put it up for review anyway to see what everyone thinks.

If this is the approach we end up using, it's going to require extensive snapshot testing, and we should probably also bump the version number to 1.1.0.


srs-bsns:
FYI: This will revert a2fb7b82fface7096c1e652a6279c7df81917ffa which fixed #392, however, the issue no longer seems relevant.

Zeno410:
Generally having the event handlers as separate objects is a good thing. There is a problem with registering on world load, though, at least in 1.7.10: some generation is done in the instantiation of the world object, before it gets loaded (I have had many tears over this insane design decision). Also, biome gen init occurs earlier, so if we don't stick to our stringent "don't futz with the biome generator" policy that might cause problems too.

srs-bsns:
> There is a problem with registering on world load, though, at least in 1.7.10: some generation is done in the instantiation of the world object

Yeah, that crap had me pulling my hair out for a while. I couldn't figure out why a certain event wasn't firing when the event class was clearly being registered. It wasn't until later that I realised I had to register the event at an earlier point in order for it to actually receive the events it needed to do it's thing. The event system isn't very intuitive at all.

PR #836, opened 2016-07-16
Ported 1.7.10 event manager stuff.

whichonespink44:
@srs-bsns Would you mind casting an eye over this before I merge into dev?

### DIFF
--- a/src/main/java/teamrtg/rtg/core/RTG.java
+++ b/src/main/java/teamrtg/rtg/core/RTG.java
@@ -1,5 +1,8 @@
 package teamrtg.rtg.core;
 
+import java.util.ArrayList;
+
+import net.minecraft.world.gen.structure.MapGenStructureIO;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.fml.common.Mod;
 import net.minecraftforge.fml.common.Mod.EventHandler;
@@ -8,15 +11,18 @@ import net.minecraftforge.fml.common.SidedProxy;
 import net.minecraftforge.fml.common.event.FMLInitializationEvent;
 import net.minecraftforge.fml.common.event.FMLPostInitializationEvent;
 import net.minecraftforge.fml.common.event.FMLPreInitializationEvent;
+import net.minecraftforge.fml.common.event.FMLServerStoppedEvent;
 import net.minecraftforge.fml.relauncher.Side;
 import teamrtg.rtg.api.module.Mods;
 import teamrtg.rtg.api.util.RealisticBiomePresenceTester;
+import teamrtg.rtg.api.util.debug.Logger;
+import teamrtg.rtg.api.world.RealisticBiomeFaker;
 import teamrtg.rtg.client.DebugHandler;
 import teamrtg.rtg.core.event.EventManagerRTG;
 import teamrtg.rtg.core.world.WorldTypeRTG;
-import teamrtg.rtg.api.world.RealisticBiomeFaker;
-
-import java.util.ArrayList;
+import teamrtg.rtg.core.world.gen.structure.MapGenScatteredFeatureRTG;
+import teamrtg.rtg.core.world.gen.structure.MapGenVillageRTG;
+import teamrtg.rtg.core.world.gen.structure.StructureOceanMonumentRTG;
 
 @Mod(modid = ModInfo.MOD_ID, name = ModInfo.MOD_NAME, version = ModInfo.MOD_VERSION, dependencies = "required-after:Forge@[" + ModInfo.FORGE_DEP + ",)" + ModInfo.MOD_DEPS, acceptableRemoteVersions = "*")
 public class RTG {
@@ -38,11 +44,12 @@ public class RTG {
 
         configPath = event.getModConfigurationDirectory() + "/RTG/";
         Mods.syncAllConfigs();
+        
+        // This must come before the event manager is registered.
+        this.registerStructures();
 
+        Logger.info("[FMLPreInitializationEvent] Creating RTG's EventManager");
         eventMgr = new EventManagerRTG();
-        MinecraftForge.EVENT_BUS.register(eventMgr);
-        MinecraftForge.ORE_GEN_BUS.register(eventMgr);
-        MinecraftForge.TERRAIN_GEN_BUS.register(eventMgr);
 
         worldtype = new WorldTypeRTG("RTG");
     }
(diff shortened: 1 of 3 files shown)


src/main/java/teamrtg/rtg/core/RTG.java
```diff
@@ -38,11 +44,12 @@ public void preInit(FMLPreInitializationEvent event) {
 
         configPath = event.getModConfigurationDirectory() + "/RTG/";
         Mods.syncAllConfigs();
+        
+        // This must come before the event manager is registered.
+        this.registerStructures();
```
srs-bsns, line 50: 1. Why move the registrations to a method?
2. What specific event(s) does the registration have to happen before? and why?

Being that the world gen specific events don't get registered until the server is going to start up, we can probably register these at anytime during the pre-init or init process. (I had originally moved the registrations in 1.7.10 to post-init and I don't recall noticing any side effects.)

src/main/java/teamrtg/rtg/core/RTG.java
```diff
@@ -38,11 +44,12 @@ public void preInit(FMLPreInitializationEvent event) {
...
 
+        Logger.info("[FMLPreInitializationEvent] Creating RTG's EventManager");
         eventMgr = new EventManagerRTG();
-        MinecraftForge.EVENT_BUS.register(eventMgr);
-        MinecraftForge.ORE_GEN_BUS.register(eventMgr);
-        MinecraftForge.TERRAIN_GEN_BUS.register(eventMgr);
 
         worldtype = new WorldTypeRTG("RTG");
```
srs-bsns: Meh.. I need to redo this again.. I swear I've fixed this like 10 times now and everytime I end up not getting the commit pushed.

`worldtype = new WorldTypeRTG(ModInfo.MOD_ID)`

src/main/java/teamrtg/rtg/core/event/EventManagerRTG.java
```diff
@@ -33,238 +35,349 @@
...
+{
 
+    private final LoadChunkRTG LOAD_CHUNK_RTG = new LoadChunkRTG();
+    private final GenerateMinableRTG GENERATE_MINABLE_RTG = new GenerateMinableRTG();
+    private final InitBiomeGensRTG INIT_BIOME_GENS_RTG = new InitBiomeGensRTG();
+    private final InitMapGenRTG INIT_MAP_GEN_RTG = new InitMapGenRTG();
+    private final SaplingGrowTreeRTG SAPLING_GROW_TREE_RTG = new SaplingGrowTreeRTG();
+    
+    public final WorldEventRTG WORLD_EVENT_RTG = new WorldEventRTG();
+    public final RTGEventRegister RTG_EVENT_REGISTER = new RTGEventRegister();
```
srs-bsns: These can be made private.

src/main/java/teamrtg/rtg/core/event/EventManagerRTG.java
```diff
@@ -33,238 +35,349 @@
...
+
+    private WeakHashMap<Integer, Acceptor<ChunkEvent.Load>> chunkLoadEvents = new WeakHashMap<>();
+    private boolean registered = false;
+    private long worldSeed;
+    private final String EVENT_SYSTEM = "RTG Event System: ";
```
srs-bsns, line 52: I don't understand the reason for storing a static string in a field in this manner. It's not a complex string or anything.. you could have just copy&pasted it in the very few instances it is used in this class, instead of creating another field that just takes up more memory. :P

src/main/java/teamrtg/rtg/core/event/EventManagerRTG.java
```diff
@@ -33,238 +35,349 @@
...
+                }
+            } else if (event.getType() == InitMapGenEvent.EventType.RAVINE) {
+
+                if (Mods.RTG.config.ENABLE_RAVINE_MODIFICATIONS.get()) {
+
+                    event.setNewGen(new MapGenRavineRTG());
+                }
+            } else if (event.getType() == InitMapGenEvent.EventType.OCEAN_MONUMENT) {
+                event.setNewGen(new StructureOceanMonumentRTG());
             }
```
srs-bsns, line 209: This would probably be better organised and more efficient as a switch statement as I did in 1.7.10: [EventManagerRTG.java#L148-L173](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/cbc148503775c972b6599e78b95b9a18af8999db/src/main/java/rtg/event/EventManagerRTG.java#L148-L173)
Note: The missing cases can also be added and left blank for future reference if we ever need to do anything with other InitMapGenEvent events.

src/main/java/teamrtg/rtg/core/RTG.java
whichonespink44, line 50: I guess I was just trying to keep the preInit a bit cleaner, but happy to move them back.

They might not _need_ to come before the event manager is registered, but I think they _should_ because InitMapGen references them, and in general the sooner they're registered the better in case any other mods depend on them being available early.

src/main/java/teamrtg/rtg/core/RTG.java
whichonespink44: I'll fix it when I'm resolving conflicts.

src/main/java/teamrtg/rtg/core/event/EventManagerRTG.java
whichonespink44, line 52: If we decide to change the prefix, then we only need to change it one place. With all the crazy decoration, surfacing and terrain generation we're doing, I think we can afford to create an extra one-off string ;)

src/main/java/teamrtg/rtg/core/RTG.java
srs-bsns, line 50: > I guess I was just trying to keep the preInit a bit cleaner, but happy to move them back.

Doesn't really matter too much, but if you're going to go through the effort to put that code into a separate method to keep the main class cleaner, then you may as well put it into static methods in the related classes, and call the method there during pre-init, that way the main class only has the method calls. :)

src/main/java/teamrtg/rtg/core/RTG.java
srs-bsns: \o/

src/main/java/teamrtg/rtg/core/event/EventManagerRTG.java
srs-bsns, line 52: Sure, if you want. Personally I consider such things 'clutter.' ;P

PR #837, opened 2016-07-16
Disabled delayed decorations and cleaned up code

topisani:
I have been getting rid of some warnings and removing redundant code. Also i commented out the delayed deco stuff in CPRTG again, so this could be released

### DIFF
--- a/src/main/java/teamrtg/rtg/core/RTG.java
+++ b/src/main/java/teamrtg/rtg/core/RTG.java
@@ -30,6 +30,7 @@ public class RTG {
     @Instance(ModInfo.MOD_ID)
     public static RTG instance;
     public static String configPath;
+    @SuppressWarnings("WeakerAccess")
     public static WorldTypeRTG worldtype;
     public static EventManagerRTG eventMgr;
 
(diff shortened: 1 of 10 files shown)


src/main/java/teamrtg/rtg/core/world/WorldTypeRTG.java
```diff
@@ -9,7 +9,7 @@
 
 public class WorldTypeRTG extends WorldType {
 
-    public WorldTypeRTG(String name) {
+    public WorldTypeRTG() {
         super("RTG");
```
srs-bsns, line 13: carl-facepalm

src/main/java/teamrtg/rtg/core/RTG.java
```diff
@@ -30,6 +30,7 @@
     @Instance(ModInfo.MOD_ID)
     public static RTG instance;
     public static String configPath;
+    @SuppressWarnings("WeakerAccess")
```
srs-bsns, line 33: This would be better suppressing the whole class, and not just that field.

src/main/java/teamrtg/rtg/core/RTG.java
```diff
@@ -52,7 +53,19 @@ public void preInit(FMLPreInitializationEvent event) {
...
+        // Scattered features
+        MapGenStructureIO.registerStructure(MapGenScatteredFeatureRTG.Start.class, "rtg_MapGenScatteredFeatureRTG");
+
+        // Villages
+        if (Mods.RTG.config.ENABLE_VILLAGE_MODIFICATIONS.get())
+            MapGenStructureIO.registerStructure(MapGenVillageRTG.Start.class, "rtg_MapGenVillageRTG");
+
+        // Ocean monuments
+        MapGenStructureIO.registerStructure(StructureOceanMonumentRTG.StartMonument.class, "rtg_MapGenOceanMonumentRTG");
     }
```
srs-bsns, line 69: This was an unnecessary code restructuring. This method was fine where it was. In fact, if you were going to move this method here then you may as well have just remove the method entirely and put those 3 registration lines back into `preInit()`

PR #854, opened 2016-07-22
Surface wrapping

whichonespink44:
So this is where I'm at so far with the surface wrapping.

Couple of things worth mentioning at this stage:
- Not all of the 1.7.10 BOP/AC surfaces have been ported across yet (just using the generic surface template for now)
- Most of the surface wrappers in SurfaceBase have placeholder names, based on the biome family (or individual biome) they came from. This is only temporary - ideally, the names should describe the surfaces themselves, not the biomes. This will change eventually.
- There are probably some 'duplicate' surface wrappers in SurfaceBase, so I still need to go through and see where I've doubled-up.


whichonespink44:
Merging this into dev before it gets too stale. We can always refactor the surface wrapper names & port the rest of the 1.7.10 surfaces later.

PR #856, opened 2016-07-24
Added biome support for FlowerCraft.

whichonespink44:


src/main/java/rtg/RTG.java
```diff
@@ -33,9 +22,11 @@
 import rtg.world.biome.realistic.abyssalcraft.RealisticBiomeACBase;
 import rtg.world.biome.realistic.biomesoplenty.RealisticBiomeBOPBase;
 import rtg.world.biome.realistic.buildcraft.RealisticBiomeBCBase;
+import rtg.world.biome.realistic.flowercraft.RealisticBiomeFCBase;
 import rtg.world.biome.realistic.minestrappolation.RealisticBiomeMSBase;
 import rtg.world.biome.realistic.thaumcraft.RealisticBiomeTCBase;
 import rtg.world.biome.realistic.vanilla.RealisticBiomeVanillaBase;
+import static rtg.reference.ModInfo.*;
```
srs-bsns, line 29: Instead of using a static import, using the fully qualified name would probably be more appropriate in this instance (ModInfo.MOD_ID, etc..) This is one of the reasons why ModInfo was kept in the root package, so that it can be called quite easily, anywhere.

PR #859, opened 2016-07-24
Added world type notification screen.

whichonespink44:
This displays a notification screen when creating a new world (similar to the one that BOP uses) which informs the user that the 'Realistic' world type must be selected in order to get realistic terrain.

Hopefully this will help prevent people from choosing the 'Default' world type and thinking that they're seeing 'realistic' terrain.

Couple of things worth mentioning...
- The two main files where all the magic happens were lifted from BOP and tweaked as needed. I asked @Adubbz a few months ago on the BOP Discord server about adding something like this to RTG, and he pointed me in the direction of BOP's gui screen, so hopefully that's all the permission we need to use this. (Adubbz, if you're not cool with this for any reason, please let me know.)
- The event handler is its own class and is registered independently of the rest of RTG's other event handlers. This is something we should probably refactor, and I probably will before we release 1.1.0 tonight.
- The localisation files are all in English... not sure how we go about translating them. We could try Google Translate, though I think we'd probably have a hard time determining where to break up the lines.


topisani:
I can do danish and german, the german one is probably useful to some people. @VelocityRa should do a greek one, and we can probably ask our users for the rest.

srs-bsns:
I might add en_PT if I get bored.

srs-bsns:
Why was 9ff1ad43982f796f13d7cd60aaa4e6e82c052d53 in this PR?  :neutral_face:

whichonespink44:
I was waitin for Adubbz to get back to me before merging this into dev and couldn't be bothered to make another branch :blush:

srs-bsns:
slacker

whichonespink44:
... out of a DeLorean?!!?!

PR #873, opened 2016-07-28
Tweaked the event management system.

whichonespink44:
This should be working for all scenarios.

I've tested all combinations of loading and unloading RTG and non-RTG (superflat) worlds, and I didn't experience any crashes, chunk walls, or stone rivers.

But it's late, and this could do with more testing... but if everything checks out, we can release 1.1.1 ~~tomorrow night~~ later tonight.

### DIFF
--- a/src/main/java/rtg/event/EventManagerRTG.java
+++ b/src/main/java/rtg/event/EventManagerRTG.java
@@ -1,12 +1,14 @@
 package rtg.event;
 
-import cpw.mods.fml.common.eventhandler.Event.Result;
-import cpw.mods.fml.common.eventhandler.EventPriority;
-import cpw.mods.fml.common.eventhandler.SubscribeEvent;
+import java.util.ArrayList;
+import java.util.Random;
+import java.util.WeakHashMap;
+
 import net.minecraft.block.Block;
 import net.minecraft.init.Blocks;
 import net.minecraft.world.World;
 import net.minecraft.world.biome.BiomeGenBase;
+
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.terraingen.InitMapGenEvent;
 import net.minecraftforge.event.terraingen.OreGenEvent;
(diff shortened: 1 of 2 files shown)


src/main/java/rtg/event/EventManagerRTG.java
```diff
@@ -107,22 +104,27 @@ public void generateMinableRTG(OreGenEvent.GenerateMinable event) {
...
+        InitBiomeGensRTG() {
+            logEventMessage("Initialising InitBiomeGensRTG...");
         }
 
         @SubscribeEvent
         public void initBiomeGensRTG(WorldTypeEvent.InitBiomeGens event) {
 
-            if (event.newBiomeGens[0].getClass().getName().contains("GenLayerEB")) return;
+            if (!(event.worldType instanceof WorldTypeRTG)) {
+                return;
```
srs-bsns, line 131: @whichonespink44 
Because you are now registering all of the event handlers during pre-init, they will all be registered at this point if you are creating a world that is not a RTG WorldType.

`return`ing here only prevents the river remover from being initialised, but we also need to unregister our other event handlers so that we don't interfere in a non-RTG world.

``` java
    if (!(event.worldType instanceof WorldTypeRTG)) {
        unRegisterEventHandlers();
        return;
    }
```

srs-bsns:
@whichonespink44
This will effectively go back your original functionality from the first PR to fix the Event Manager. If RTG is loaded, our event handlers are all registered, and are only unregistered if creating/loading a world that is not WorldTypeRTG, and then reregistered upon world unloading. This is probably for the best.

When I went with the opposite functionality of ONLY registering our event handlers when it was WorldTypeRTG and not until a world was about to be created/loaded, I was concerned with possible compatibility conflicts between the time minecraft loads to when a world is loaded, but I was probably just being overly cautious.

src/main/java/rtg/event/EventManagerRTG.java
whichonespink44, line 131: I don't think any of RTG's other event handlers need to be unregistered this early, as they'll all get unregistered before the non-RTG world loads (WorldEvent.Load), but... like I said in the comments, it's better to be safe than sorry, so unregistering them here as well doesn't hurt anything.

whichonespink44:
Ok, I've been testing this on and off all day, and I've been unable to find anything wrong. Works fine in superflat & BOP worlds and any combination of logging in/out between those worlds and RTG worlds seems to be working fine. Only thing I haven't tested is SMP, which I'll be doing when I get home.

If anyone can think of any other combos/scenarios I might've missed, let me know or give it a whirl and post any issues here.

whichonespink44:
Merging this before going down the 'always-registered, dim0-checking' approach as discussed in Discord.

PR #881, opened 2016-07-29
Added Hungarian translation

kesu1:


whichonespink44:
@kesu1 Thank you very much for the contribution! :)

PR #921, opened 2016-08-23
World type screen only displays if enabled in configs.

whichonespink44:


srs-bsns:
where does this get set to `false` after it's been displayed once?

whichonespink44:
It doesn't - I just added a config option that skips the registration of WorldTypeMessageEventHandler as I didn't feel like spending too much time on something that's already working.

PR #928, opened 2016-08-28
Fixed flowing liquids.

whichonespink44:
Right, so... where to begin!

Basically, in 1.10.2 flowing liquids (waterfalls/lavafalls) are handled by the base biome's BiomeDecorator instead of the ChunkProvider, which was effectively causing flowing liquids to generate whenever DecoBaseBiomeDecorations was being called (because that calls the base biome's decorate() method, which uses BiomeDecorator), even when flowing liquids were disabled in the configs.

So... in order to get around this, I had to add an event handler to DecorateBiomeEventRTG which cancels vanilla flowing liquid generation so that RTG can handle it.

While I was figuring all this out, I spent a lot of time in CPRTG and decided to move some of the decoration-related stuff out of CPRTG and into the new BiomeDecoratorRTG, which ended up being just ore gen and clay gen because I didn't want to mess with the delayed decoration system. I'm probably going to move rPopulatePreDecorate() and rPopulatePostDecorate() out of RBB and into BiomeDecoratorRTG as well since those aren't doing anything major.

So I still need to finish up some stuff, including adding a config option to enable/disable the event cancelling, but I'm putting this up for review now, just in case anyone wants to take a look and see if they can spot anything dodgy. I know it probably seems like a lot of change for something as trivial as flowing liquids, but... we're still in snapshot phase, and preliminary tests have gone well as far as I can tell, so... I think we'll be alright.


whichonespink44:
Ok, I've been doing more testing, and everything seems to be working fine. Gonna merge this now before it goes too far off-scope.

PR #942, opened 2016-09-13
Added biome support for Agricultural Revolution.

whichonespink44:
This is ready to go, except for one major show-stopper: all of AR's oceanic biomes crash when trying to generate:

http://pastebin.com/qquQtyrB

I've tried adding a try/catch to the biome's call to DecoBaseBiomeDecorations, but the crash is happening before then, during chunk population.

Agricultural Revolution is closed source, but using JD-GUI, I inspected the code but didn't spot anything that could be causing the crash. However, there's a lot of code, and without a global search function, it was difficult to see how everything fits together.

All other world gen (including world gen that spawns in vanilla Ocean and Deep Ocean) is generating fine, so I think it must be crashing when trying to generate structures in AR's oceanic biomes.

I'm posting this here in case anyone wants to take it for a test drive. Everything else is ready to go - just need to get to the bottom of the oceanic biome crashes.

Worst-case scenario, we could always just _not_ support those oceanic biomes, but I'd like to get to the bottom of these crashes at some point. (This is OceanCraft all over again!) :sob:


whichonespink44:
Turns out the reason the oceanic biomes were crashing is because I was using Lonely Biome to test, which was causing infinite decoration loops in AR.

Using a seed that mr_moeman supplied, I was able find all 3 oceanic biomes in the wild, and they're all generating fine in RTG worlds, so I'll merge this after I merge the Forge update PR.

PR #949, opened 2016-09-16
Fix ModInfo, clean up mod container class, update build scripts

srs-bsns:


whichonespink44:
Should this be targeting the 'flippage' branch instead? I can re-target it if so.

srs-bsns:
It was supposed to be from srs-bsns/1.10.2-dev to Team-RTG/1.10.2-dev, but I didn't realise at the time I was still on the flippage branch locally.
But yeah, it should be retargetted, and this PR's commits can just get merged from the flippage branch later. It's not completely ideal, but there's no priority on this PR anyways.

PR #952, opened 2016-09-17
The Great Unflipping Campaign of 2016

whichonespink44:
Since this is almost finished, I thought I'd create the PR in case anyone wants to take it for a test drive.

The only issue remaining is the weird ice flipping, but other than that, things seem to be back to normal.


whichonespink44:
Cheers!!

PR #968, opened 2016-09-25
Better structure management.

whichonespink44:
Ok, so... it turns out we don't even need InitMapGenRTG since we're using our own chunk provider, which means we can get rid of our preBiomeDecorate() event handler as well, which was only there so we could check to see if we're in an RTG world for the map gen event.

Since structures get assigned during chunk provision, and we already have access to the world at that stage, all we need to do is use our structure code if we're in an RTG world (and modifications have been enabled) and vanilla structure code if we're not. We're still posting to the TerrainGen.getModdedMapGen event, so other structure mods can still do their thing, but we're not listening to that event ourselves anymore, because we don't need to.

The only side effect of this is... if 'enableOceanMonumentModifications' is set to FALSE, then ocean monuments won't generate in RTG worlds because the vanilla ocean monument code doesn't work in RTG worlds. I haven't tested what happens if the other 'enabledXXXModifications' options are set to false, but all of the others _should_ work.

I'm submitting this PR for review now so that I can get some feedback whilst I'm testing the other modification options.

### DIFF
--- a/etc/config/RTG/rtg.cfg
+++ b/etc/config/RTG/rtg.cfg
@@ -56,6 +56,7 @@ caves {
 
     # Must be set to TRUE for the other cave settings to have any effect.
     # If FALSE, RTG won't interfere with cave generation at all.
+    # WARNING! Setting this to FALSE may result in unpredictable cave generation.
     #  [default: true]
     B:"Enable Cave Modifications"=true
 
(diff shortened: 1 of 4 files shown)


whichonespink44:
Looks like we need to remove the option to disable ocean monument modifications... if we let RTG use vanilla's code, they generate where they're not supposed to instead of not generating at all.

I've confirmed that all other structures are generating with all of the 'enabledXXXModifications' options set to false.

PR #973, opened 2016-09-28
BetterAgriculture support

lightningo7:
Adds support for the "BetterAgriculture" mod and it's "farmlandBiome" biome. Also adds wheat generator to the world.


whichonespink44:
Nice!

I'll take this for a test drive tonight, but looks good code-wise.

PR #975, opened 2016-09-29
Some clean-up

srs-bsns:
- update build scripts
  -- proper dependencies, remove third-party source
  -- added gradle property for verbose warnings
- fixed ModInfo.java
  -- proper version ranging for Forge
  -- readded mod dependencies
- updated mod_deps.properties
- cleaned up RTG.java
  -- updated container annotation
  -- fixed ambiguity
  -- moved some things for better clarity


build.gradle
```diff
@@ -26,28 +26,35 @@ ext.ref = parseprops(file('build.properties'))
...
 
 group = ref.package_base + '.' + ref.mod_id
-archivesBaseName = ref.mod_abbr
-// remove travis stuff from version for now because we aren't using it and it was a mess. -srs
-version = ref.mc_version + '-' + ref.mod_version
+archivesBaseName = (ref.mod_id as String).toUpperCase() + '-' + (ref.mc_version as String)
+version = ref.mod_version
+
+repositories {
+    maven { url "http://files.minecraftforge.net/maven" } //BiomesOPlenty
```
whichonespink44, line 33: If we add these Maven repos, does that mean we can get rid of the BOP and AC APIs in src/api?

build.properties
```diff
@@ -10,10 +9,12 @@ mod_url=https://github.com/Team-RTG/Realistic-Terrain-Generation
 mod_author="Team RTG"
 mod_creds=Forked from RWG by ted80. OpenSimplex Noise by KdotJPG.
 mod_logo=assets/rtg/logo.png
-mcf_version=12.18.1.2011
-# mcf_postfix is the branch postfix (without '-') on the Forge version when it's not the default branch in the Forge repo
-# This should be same as mc_version, eg Non-default: 1.9.4-12.17.0.1974-1.9.4, Default: 1.9.4-12.17.0.1976
-mcf_postfix=
-mappings=snapshot_nodoc_20160724
+mcf_version=12.18.1.2094
```
whichonespink44, line 12: I'm not sure we should be updating Forge when we're 3 days from release. I also don't think we should be developing on a non-recommended version of Forge. Can we change this back to 2011 and revert the mappings?

mod_deps.properties
```diff
@@ -1,8 +1,9 @@
...
-depstring=after:BiomesOPlenty@[5.0.0.2067,);\
-            after:AbyssalCraftAPI@[1.8,);\
+bopver=1.10.2-5.0.0.2085
+acver=1.10.2-1.9.2.9
+depstring=after:BiomesOPlenty@[5.0.0.2085,);\
+            after:AbyssalCraftAPI@[1.8.1,);\
             after:BiomesYouGo@[3.0.2,);\
-            after:flowercraft;\
-            after:sugiforest
+            after:flowercraft@[6.01,);\
```
whichonespink44, line 7: There was a reason I didn't add a version string to Flowercraft and Sugiforest, although I can't remember what it was! 😜 I'll need to refresh my memory, but I think it had something to do with dodgy version numbers in the JAR not matching up or something.

whichonespink44 (changes requested):
@srs-bsns I was going to just chat about this stuff with you in Discord, but I've been meaning to try out this new review system for a while, so... thought I'd do it here instead.

mod_deps.properties
whichonespink44: Flowercraft is fine... author must've updated the mod ID (or I was just having a momentary lapse of reason when I was adding biome support).

build.gradle
srs-bsns: ~~Dammit, that was supposed to be part of this commit. Will add it in a bit, so don't merge it yet.~~
Commit added.

build.properties
srs-bsns: I've already been using this version of Forge and haven't noticed anything wrong. None of the changes made since 2011 have any effect on world gen that isn't backwards compatible. The min version is still set to 2011, so no one is forced to use the latest. The mappings don't make a difference as it only effects the deobf dev environment.

mod_deps.properties
srs-bsns: One of those mods wasn't setting the version in their mod container class, but in that case it will be set from the mcmod.info, which had a version set.
I never tested this, but I don't see any reason why there would be a problem here.

whichonespink44: Ahh yes, I was just erring on the side of caution cuz I didn't know how it worked and didn't test it like I should have. Just tested it now though, and Sugiforest loaded up no problem.

build.properties
whichonespink44: I noticed quite a few commits since 2011 that _could_ affect RTG:
- [Fix async race condition on new chunks](https://github.com/MinecraftForge/MinecraftForge/commit/cc0e6a182568a282c1daa61ad300c0d29d422f45)
- [Update decorate for biomes to support Decorate event](https://github.com/MinecraftForge/MinecraftForge/commit/0cf2cdfdb3fa603379d165c2e46cb0b57383501d)
- [Fix a vanilla bug with the blockstate ID map not being properly symmetric with respect to block.getStateFromMeta](https://github.com/MinecraftForge/MinecraftForge/commit/ab64d1a1665e93b3d1fa46065119356eabe7efd3)
- [Fix vanilla bug in BlockFalling that caused blocks to loose state info during world gen](https://github.com/MinecraftForge/MinecraftForge/commit/b5db198f8b97618febd7a48a57cd4f64fdd23e25)

Whilst most of those _should_ be changes for the better, we don't know for certain that they won't break something in RTG, or introduce some weird incompatibility involving RTG and some other mod.

I also really, _really_ don't like building against a non-recommended version of Forge. It's recommended for a reason - it's supposed to serve as a common playing field for developers and users alike. By building against a non-recommended version of Forge, we're basically throwing all of that out the window.

I don't know... am I making mountains out of molehills here? Every instinct I have is telling me that this is a bad idea when we're 3 days from release, but... does it really matter? What's the worst that could happen... people start posting tons of crash reports due to some obscure incompatibility with the latest Forge, so we revert back to the recommended version of Forge and release 4.1.1.3 the next day?

build.gradle
```diff
@@ -26,28 +26,35 @@ ext.ref = parseprops(file('build.properties'))
...
     makeObfSourceJar = false
-    replace '@MOD_NAME@', ref.mod_name
-    replace '@MOD_ID@', ref.mod_id
-    replace '@MOD_VERSION@', version
-    replace '@MCF_VERSION@', ref.mcf_version
-    if (ref.mod_deps!="") replace '@MOD_DEPS@', ";" + ref.mod_deps.depstring
+    replace '@MOD_ID@',         ref.mod_id
+    replace '@MOD_NAME@',       ref.mod_name
+    replace '@MOD_VERSION@',    project.version
+    replace '@MCF_MINVER@',     ref.mcf_minver
```
whichonespink44, line 51: I'm getting a crash when trying to launch MC from within IDEA:

`java.lang.RuntimeException: Invalid versioned reference Forge@[@MCF_MINVER@,@MCF_MAXVER@)@MOD_DEPS@`

Anything we can do?

whichonespink44 (changes requested):
@srs-bsns I know you mentioned something about using a different placeholder that would allow us to get around this issue, but can't remember what we said we'd use.

build.properties
srs-bsns: The thing is, most users are going to be using the `latest` Forge, or atleast something newer than the `recommaneded`. It's best if we build against the `latest` to ensure compatibility with what users are using. If there are any incompatibilities we'll likely catch them in development, or initial testing. If we build against `recommended` and most users are using `latest` and run into issues, it's going to be a headache to track something down that is related to a Forge update.

Also, the Forge devs are quite anal about introducing changes that would break compatibility. If something was introduced that would break old versions there would likely be warnings issued. Things that could possibly break compatibility are usually reserved for major updates (ie. waiting until 1.11 before enforcing lowercase-only mod IDs)

After watching Forge for a while now, it seems the 'recomended' version is set more by guildline than by rule. In the past there's been versions set to 'recommended' directly after code has been committed and not really tested (publically), and other times, when the recommended version should be bumped, it isn't. (ie The recommended version for 1.7.10 is still 1566, but could have been bumped to 1614 a year ago, since it's the version almost everyone uses and has been thoroughly tested by now.)

#### As for those commits:

> Fix async race condition on new chunks

This shouldn't affect anything in RTG. The only thing it would have affected is the village code that we've already removed, but this would have fixed things, rather than break them. (No more broken buildings in villages, and possibly no more random undecorated chunks.)

> Update decorate for biomes to support Decorate event

This updates biome decoration events to add support for 1.10 added stuff, which shouldn't affect anything we're doing, though in the future we should update to add support for fossils, etc. This should be transparent to RTG.

> Fix a vanilla bug with the blockstate ID map not being properly symmetric with respect to block.getStateFromMeta

This seems irrelevant, since `Block#getStateFromMeta` is deprecated and just returns `Block#getDefaultState` regardless of the meta value passed. Everything has moved to properties. I'm not sure the purpose why this was even commited to begin with, though when this patch was initially being written it was for MC 1.8. This shouldn't have any direct effect on RTG.

> Fix vanilla bug in BlockFalling that caused blocks to loose state info during world gen

This is an OLD vanilla bug. It shouldn't affect anything that we do, but vanilla blocks that can fall shouldn't remain suspended anymore with no solid blocks below them. This should be completely transparent to RTG.

---

If you're still worried about it, you can push the version back, or we can do another snapshot after building against the `latest` Forge and wait another week until release. I know the plan was to release this weekend, and there are users looking for a stable RTG release, but that shouldn't be a reason to rush it.

build.gradle
srs-bsns: I'll have a look at this now, and see what I can do.

build.properties
keybounce: > This seems irrelevant, since Block#getStateFromMeta is deprecated and just returns Block#getDefaultState regardless of the meta value passed.

I was under the impression that you are supposed to override Block#getStateFromMeta(), and not rely on that default implementation for anything -- i.e, it is only there to make sure that something non-crashing happens.

srs-bsns: I don't think you understand [software deprecation](https://en.wikipedia.org/wiki/Deprecated#Software_deprecation).

whichonespink44:
@srs-bsns Since this PR has conflicts, do you want to create a new PR with just the new mod deps, Forge update, and new mappings? Or I can just copy the relevant bits from here if you want? Or just update from dev so that this isn't conflicted anymore? I'm easy either way.

Still need to have a think about what the best way is to include the API's from other mods that won't affect local testing. I know you posted a potential solution in Discord, so hopefully that'll do the trick, but if not, then I'm happy to keep API's in src/api as there's really no reason to include the entire deobj library as part of the gradle project when we only need the API. (Or is it just the API deobf that gets included?)

build.properties
keybounce: Normally, it means that something is being phased out because there's a better way to do it.

But: To write a block to the disk, you have to convert a state to a 4 bit meta. Reverse that on the way in. So what has replaced getStateFromMeta and getMetaFromState?

PR #980, opened 2016-10-03
Improve generic block crops world gen for determining maximum growth.

gabizou:
This is a pretty straightforward cleaner way to retrieve the maximum growth for `WorldGenCrops` and determine the max growth `IBlockState`. I've tested it locally and it works in vanilla environments (Which is what I understand this to be targeting in the first place).

Let me know if this should be changed at all or anything.


whichonespink44:
Thanks @gabizou! :)

PR #983, opened 2016-10-03
getStateFromMeta to Properties

lightningo7:
Edited most of the getStateFromMeta appearances to start using Properties.

Also a few minor OCD things.


whichonespink44:
Just checking this out now... if there are any changes that need to be made, I'll try another one of these GitHub review things.

whichonespink44:
Moved to #984

PR #984, opened 2016-10-03
getStateFromMeta to Properties

whichonespink44:
@lightningo7 Everything that I've looked at looks good, though I haven't checked absolutely everything.

I've created this remote feature branch so that I could push some changes.

I've added a new class called BlockUtil, which we can use for the more complicated getStateFromMeta calls... the structure's there to support whatever blocks we need to, but I've only implemented the sand block for now.

I can either merge this as-is right now, or you/we can try to go through the rest of the getStateFromMeta occurrences and use BlockUtil.getState() instead of repeating all the same code over and over.

Happy to merge it now though if you want, I'm easy either way.


whichonespink44:
@lightningo7 By the way, if you want to push changes to this PR, you'll need to target the `pr/983` branch when submitting the PR.

src/main/java/rtg/util/BlockUtil.java
```diff
@@ -0,0 +1,38 @@
...
+    public static IBlockState getState(Block block, int meta) {
+
+        if (block == Blocks.SAND) {
+            switch (meta) {
+                case 0:
+                    return Blocks.SAND.getDefaultState();
+                case 1:
+                    return Blocks.SAND.getStateFromMeta(BlockSand.EnumType.RED_SAND.getMetadata());
+                default:
+                    return block.getStateFromMeta(meta);
```
lightningo7: I don't quite understand why we're using getStateFromMeta in this case? Aren't we trying to remove these occurrences with this PR? It could just be me but is it not inefficient to be sending it through a massive else if statement each time it requests a block? Would it not be faster to do each one individually? Though I will be going through all the other occurrences and replacing them pointing them to this file.

whichonespink44: There might be a better way to do it, but I think the main benefit of running the instances of getStateFromMeta which involve some unknown meta value is that you don't have to use the same code over and over again to determine which blockstate to cough up.. all of that logic can be contained in the BlockUtil class, which is really just a stop-gap measure until we remove the use of metas from RTG completely (e.g. the config system).

But if you can fix each one individually without writing the same code over and over again to determine which state to use, then go for it. BlockUtil is just there if you need it. And no, it's not ideal running everything through a big if/else statement, but it's got nothing to do with performance. That's not gonna affect anything significantly. But there's probably a better way to do it for sure. Again, it's just there if you need it. Feel free to tweak it to make it better if you want.

As for the getStateFromMeta() for the sand block... the thing is... getStateFromMeta isn't _really_ deprecated... Choonster explains it well here:

> Mojang seems to be using @Deprecated to mark Block methods that should only be called by subclasses of Block, external code should be using the corresponding IBlockState methods. I've personally just put @SuppressWarnings("deprecation") on these methods in my code since they're not really deprecated.

The getStateFromMeta for Blocks.SAND is fine because Blocks.SAND overrides Block#getStateFromMeta. But for blocks that _don't_ override Block#getStateFromMeta, we should be using properties for those.

At least, that's my understanding of it.

whichonespink44:
@lightningo7 nice one... so is this everything then? If so, then I'll take it for a test drive before merging, but if you have more stuff to add, I'll hold tight.

lightningo7:
I believe this was the final batch of things to add. IF there is more, I've missed it somewhere along the line. May want to just double check for me in case. Other than that I believe it's good though.

PR #998, opened 2016-10-12
RB Terrains

whichonespink44:
Part 1 of 3 of the moving of configs/terrains/surfaces into the RealisticBiome classes.


whichonespink44:
@Team-RTG/rtg-writers 

Ok, so... I've finished moving all of the terrain classes into the biome-specific RB classes. To accomplish this, I basically just copied what @topisani did in RTG2, creating an abstract initTerrain() method in all RB classes, except instead of creating a new TerrainBase ['on-the-fly'](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/RTG2-dev/src/main/java/teamrtg/rtg/modules/vanilla/biomes/RTGBiomeVanillaForest.java#L28-L42), I'm creating an instance of the actual terrain class which I ported into the RB as an [inner class](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/e08920bd649e670b04f8fc9d39cbcd2f5cb87a89/src/main/java/rtg/world/biome/realistic/vanilla/RealisticBiomeVanillaForest.java#L31-L56).

To ensure that I ported all of the terrain classes correctly, I did some 'unit testing' by running the old & new terrain systems side-by-side and simulating some terrain gen via the [compareTerrain()](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/77d35e306e3c1e0d4195ad09c40fa8933d0f74d3/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java#L554-L582) method that I wrote, which throws an exception if the terrains don't match up for whatever reason. It caught a few mistakes that I'd made, but after cleaning those up, I was able to successfully unit test all terrains across all of RTG's supported mods, and the new terrains are exactly the same as the old terrains.

I still need to do some in-game testing, especially on existing worlds created with 4.1.1.6, but this is basically finished and ready to be merged. I just thought I'd leave it up here for a bit in case anyone wants to take a look.

PR #1017, opened 2016-10-28
RTG trees no longer replace blocks they shouldn't.

whichonespink44:
Before an RTG tree places a block, it now checks WordGenAbstractTree#isReplaceable to see if it's allowed to do so.

As part of this PR, I've also gone through and removed all other instances of 'replaceable block checking' because it's all redundant now that we're using isReplaceable().

I'm still in the process of testing this, but I thought I'd post it for review now in case anyone spots something wrong.

### DIFF
--- a/src/main/java/rtg/world/gen/feature/tree/rtg/TreeRTG.java
+++ b/src/main/java/rtg/world/gen/feature/tree/rtg/TreeRTG.java
@@ -164,4 +164,11 @@ public class TreeRTG extends WorldGenAbstractTree {
 
         return true;
     }
+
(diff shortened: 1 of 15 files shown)


whichonespink44:
Well, already noticed something wrong...

The first world I tested this in, I noticed that many of the trees' leaves are decaying immediately. 😢

whichonespink44:
Figured out why leaves were decaying - basically, wood is considered an 'isReplaceable()' block so leaves were replacing their supporting log blocks, so just created a separate method for checking leaves which only places them if an air block is found, and everything seems to be working fine. Will push latest stuff after a few more tests.

PR #1032, opened 2016-11-07
Update TreeRTG#canGrowInto to use all appropriate Materials.

### LINKED ISSUE #1031: Trees not replacing tallgrass blocks
Either trees aren't replacing tallgrass blocks, or the tallgrass is replacing log blocks. 99.99% certain it's the former but need to investigate to be certain.

It'll most likely be one of these methods:

https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/1.10.2-dev/src/main/java/rtg/world/gen/feature/tree/rtg/TreeRTG.java#L185-L211

srs-bsns:
Closes #1031


srs-bsns:
Tested and confirmed fixed.

PR #1048, opened 2016-11-13
RB Configs

whichonespink44:
Part 3 of 3 of the moving of configs/terrains/surfaces into the RealisticBiome classes.


whichonespink44:
This PR also includes a new file/folder structure for the biome configs...

Previously, RTG created one config file for each mod that it supports, regardless of whether the user had that mod installed or not.

Now, RTG creates a config **folder** for each mod, and one config file for **each biome** in that mod, but **only** if the user has the mod installed.

```
/config/RTG/biomes/vanilla/beach.cfg
/config/RTG/biomes/vanilla/birchforest.cfg
/config/RTG/biomes/vanilla/birchforesthills.cfg
/config/RTG/biomes/biomesoplenty/alps.cfg
etc.
```

PR #1055, opened 2016-11-19
update RTG to stable 1.10.2 mappings, build against Forge 12.18.2.2125

srs-bsns:
- also updated the BoP and AbyssalCraft APIs so they won't error

:heart:


whichonespink44:
Doesn't this mean that users will have to use Forge 2125 as well?

srs-bsns:
No, it only affects the dev environment. Once everything is reobfuscated it all points to the right stuff.
Which is why I left the Forge minimum version set to 2011.

srs-bsns:
Ill redo this PR at a later date

PR #1064, opened 2016-11-28
RTGWorld

whichonespink44:
This is the first round of refactoring required to implement the 'faked biome' system from v2.


whichonespink44:
The `{mod} biomes now use RTGWorld.` commit messages above should read `{mod} surfaces now use RTGWorld.`

whichonespink44:
Ok, this PR is ready to be merged into dev (after some final testing, which I'll be doing tonight).

Despite the number of changed files, nothing has changed functionality-wise.

The only major change in this PR is the introduction of RTGWorld, which gets created when CPRTG is initialized and holds an instance of World, OpenSimplexNoise, SimplexCellularNoise, and Random - all of which get used extensively throughout the mod; hence, the number of changed files. (RTGWorld will also eventually hold an instance of [RealisticBiomeFaker](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/RTG2-dev/src/main/java/teamrtg/rtg/api/world/RealisticBiomeFaker.java).)

I've also done some minor refactoring along the way.

PR #1110, opened 2017-01-14
Woodland Mansions

whichonespink44:
Here's my first attempt at woodland mansions - unfortunately, I haven't been able to get them to generate yet.

RTG tries to generate them, but that `setBlocksInChunk` method doesn't seem to be compatible with the way RTG generates the terrain, throwing lots of errors in the console (though never crashing, which I thought was weird).

The `/locate Mansion` command also doesn't seem to be working... the coords that it displays don't lead to a Roofed Forest biome, and when using a known woodland mansion seed, the mansion doesn't try to generate in the Roofed Forest biome that it's supposed to be at.

I think we need to figure out how to convert the `setBlocksInChunk` method to something compatible with RTG before we can get them working, but it involves heightmaps, which RTG handles slightly differently than vanilla, so not sure what to do here.


Ezoteric:
So adding stuff to rtg now ? make mushroom island a jungle island option :P

PR #1112, opened 2017-01-17
New X-hills, version 2

Zeno410:
For comment. Lower hills away from the ridges and wider valleys than the previous version.

### DIFF
--- a/src/main/java/rtg/RTG.java
+++ b/src/main/java/rtg/RTG.java
@@ -48,7 +48,7 @@ import static rtg.api.RTGAPI.config;
     modid                    = ModInfo.MOD_ID,
     name                     = ModInfo.MOD_NAME,
     version                  = ModInfo.MOD_VERSION,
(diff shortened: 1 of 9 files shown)


src/main/java/rtg/RTG.java
```diff
@@ -48,7 +48,7 @@
     modid                    = ModInfo.MOD_ID,
     name                     = ModInfo.MOD_NAME,
     version                  = ModInfo.MOD_VERSION,
-    dependencies             = "required-after:forge@[" + ModInfo.MCF_MINVER + "," + ModInfo.MCF_MAXVER + ")" + ModInfo.MOD_DEPS,
+    dependencies             = "required-after:Forge@[" + ModInfo.MCF_MINVER + "," + ModInfo.MCF_MAXVER + ")" + ModInfo.MOD_DEPS,
```
srs-bsns: ~~You can not do this. The Forge mod ID **has** to be all lowercase, just as with all mods, in 1.11.~~

Saw this before looking at the fallowing commit.

PR #1120, opened 2017-01-26
Volcano speed improvements (cherry-picked from 1.11)

srs-bsns:
About 50 times as fast, from about half the CPU consumption of terrain
generation to a minor effect. Also, we get to skip a lot of biome
lookups which might speed things a little more.

(cherry picked from commit 348442480fa71584b917491b40997f13cdf54c03)


srs-bsns:
@whichonespink44, Keep this open and do not merge until @Zeno410 is finished improving volcano generation in 1.11, then I'll port it all over to 1.10 and add the commits to this PR

PR #1126, opened 2017-01-28
Biome-configurable biome temperatures.

whichonespink44:
Ok, so... I got rid of the `MIN_Y_COORD_FOR_SNOW_LAYERS` *global* config option and replaced it with a `TEMPERATURE` *biome* config option.

This config option is available to all biomes, but the only biomes that RTG changes by default are the Extreme Hills variants, changing the temperature from 0.2 to 0.25, resulting in snow layers being generated at around Y=120 instead of Y=90. This works when the chunk is initially generated and also when it rains.

The magic happens during `RBB#init` via `RBB#adjustBiomeProperties`, which is the earliest that it can happen given that it's biome-configurable. However, I'm slightly concerned that it won't be early enough for mods that query biome temperature.

But I guess we can cross that bridge if we come to it?

### DIFF
--- a/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
+++ b/src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
@@ -586,10 +586,15 @@ public abstract class RealisticBiomeBase {
                 throw new RuntimeException("Biome temperature out of range for " + biomeName + ".");
             }
 
-            Accessor<Biome, Float> biomeTemp = new Accessor<>("temperature");
-            biomeTemp.setField(biome, biomeTemperature);
+            try {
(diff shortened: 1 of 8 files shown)


src/main/java/rtg/world/biome/realistic/RealisticBiomeBase.java
```diff
@@ -587,13 +587,13 @@ else if (biomeTemperature < -2f || biomeTemperature > 2f) {
             }
 
             try {
-                Accessor<Biome, Float> biomeTemp = new Accessor<>("temperature");
+                Accessor<Biome, Float> biomeTemp = new Accessor<>("temperature", "field_76750_F");
```
srs-bsns, line 590: We should really make an AT for this stuff.

PR #1153, opened 2017-03-23
Grabbag of terrain changes

Zeno410:
Revised river system and mesa plateaus. Added single-pass decoration for overdecorated Bop biomes. Changes to several Bop Biomes. Suppressed biome-specific caves. Changed voronoi basin placement algorithm and voronoi result passing. Added a couple of utilities.

PR #1169, opened 2017-05-15
Pr/1153

whichonespink44:
This is the manual re-targeting of @Zeno410 's PR #1153 (Grabbag of terrain changes) to the 1.10.2 branch.

It contains everything that's been developed for the 1.11 version, including:

**From PR #1153**
- Revised river system and mesa plateaus.
- Added single-pass decoration for overdecorated BOP biomes.
- Changes to several BOP biomes.
- Suppressed biome-specific caves & ravines.
- Changed voronoi basin placement algorithm and voronoi result passing.
- Added a couple of utilities.

**Other stuff that was originally written/fixed for 1.11.2**
- New terrain design for Extreme Hills variants, including re-surfacing.
- New terrain design for Ocean & Deep Ocean, including re-surfacing (sponge boulders!).
- Better BOP Bayous.
- Y-level tweaks for various Swamp decos.
- Re-decorated vanilla Taiga biome variants.
- Re-decorated Roofed Forest to have fewer mob spawns under canopy.
- Volcano speed improvements.
- Added TEMPERATURE biome config and set Extreme Hills variants to 0.25 by default, resulting in snow layers being placed at around Y=120 instead of Y=90.
- Various tweaks and enhancements.

@Zeno410 @srs-bsns --- I've added you guys as reviewers... could you cast your eye over this when you get some spare time and let me know if you spot anything dodgy?

If everything looks good, I'll merge this into `1.10.2-dev` and maybe even release a snapshot to get the community hyped for a new release.


whichonespink44:
And yes, I know there are some 1.11-specific commits in this PR (e.g. f3c13e5, 7378422, etc.), which isn't ideal for the commit history, but... doing it like this was the easiest way for me to get my head back into RTG again, so we're just gonna have to live with it. :)

srs-bsns:
It's been quite a while since I've looked at any of this stuff, and yes, there's a bunch of changes in f3c13e551ed8d7a420ef18a915673a06e9e0aea1 that will need to be reverted, but what about 73784222768b0cb885adda46533b799442da2623 is 1.11 specific?

Also, since this will be a major update with breaking changes, we may as well keep the lowercase Mod ID. This will better the compatibility between this next release for 1.10 and future versions.

whichonespink44:
I was just referencing some issues that were done as part of the initial update to 1.11.

Nothing to revert because it's all been reverted already :) I created a branch from Zeno's PR, then I copied the entire 'src' directory from the 1.10.2-dev branch and pasted into that branch... and then I manually went through all of the commits from the original PR and brought them across manually. There's probably a much easier way to do it, but it seems to be working fine.

As for keeping the mod ID lowercase, I'm not sure that's a good idea now. I mean, sure, the next release for 1.10.2 will world-breaking but only in the sense that it will cause chunk borders. If someone really wanted to update their 1.10.2 world to the next 1.10.2 release, they could if the mod ID was still all-caps. But they couldn't update if we change the mod ID to lowercase, even if they wanted to because that would literally break their world, right?

srs-bsns:
>because that would literally break their world, right?

No, it wouldn't, because the WorldType would still be the uppercase `RTG`, which is all that matters in that case. ;)

The only issue I think changing the Mod ID would cause is, maybe, with WTF Expedition or Recurrent Complex, because they may make checks if RTG is loaded, which would likely (properly) be done against the Mod ID. This would be a very easy fix on their side, but I'm not sure if either mod is still being maintained for 1.10 anymore. Other than that, I doubt anyone would even notice that the change has occured.

whichonespink44:
Ah, didn't know that it was the world type name that mattered, so that's cool.

But yeah, that's a good point about breaking compatibility with those mods if we change the ID.

What's the biggest benefit to changing it now rather than waiting until we release a version that requires the lowercase ID?

srs-bsns:
>What's the biggest benefit to changing it now rather than waiting until we release a version that requires the lowercase ID?

So that the compatibility of future versions will be able to extend back to any new releases for 1.10. Though, since it is the WorldType ID that is the most important factor for world compatibility, I guess a Mod ID change would be a bit trivial. I just suggested it because we're making world breaking changes anyways, but perhaps to maintain compatibility with WTF Expedition and RecCom we could keep it the way it is.

Issue #1174, opened 2017-05-21
Gaps on world generation


Scratchman:
- Forge version: forge-1.10.2-12.18.3.2297
- RTG version: RTG-1.10.2-4.2.0.0-snapshot-1

##### Issue Description:
Using your latest snapshot I've created a new world, but I'm seeing lots of "streaking"

http://map.imaginescape.tk/

##### Steps to reproduce the issue: (If it's easily reproducible.)

1. Use chunkgen-1.10.2-1.5.0 to pregenerate the world
2. render with dynmap

I can provide a list of mods in the modpack if it would help diagnose the problem

##### Extra Information:
I didn't see any errors in the server log that might explain the issue.

srs-bsns:
Not sure what's going on here yet, but clearly something is. You can see in the last screenshot that it doesn't appear to be restricted to chunk borders, as there doesn't appear to be signs of anything getting cut off at the borders, but the streak is most definately vertical in nature.

Seed: 3459487736926468020

![2017-05-21_03 18 20](https://cloud.githubusercontent.com/assets/8453133/26281968/a9887bc0-3dd4-11e7-89c3-25b3b97c206c.png)
![2017-05-21_03 19 01](https://cloud.githubusercontent.com/assets/8453133/26281971/ae04e5bc-3dd4-11e7-8ca5-863b1408e3ab.png)
![2017-05-21_03 21 06](https://cloud.githubusercontent.com/assets/8453133/26281972/b115c7c6-3dd4-11e7-9ed3-2a0d86b2af72.png)

whichonespink44:
That's the issue I saw earlier before we released the snapshot.

When I restarted MC, created a new world with the same seed, and flew to the same location, the undecorated streak was gone, so we put it down to a one-off de-sync issue.

Maybe it's related to pregenning?

Scratchman:
as you can see from the dynmap, half the map isn't generated yet, so i'll fly over that way and see if i can find any gaps.

whichonespink44:
![2017-05-21_08 43 35](https://cloud.githubusercontent.com/assets/10927161/26282082/a963401c-3e01-11e7-8696-81a59092b5d9.png)

Seed: 3459487736926468020

I didn't pre-gen anything - I just flew to the location in the screenshots above, and... no streaks.

¯\\\_(ツ)_/¯

Scratchman:
[...]

PR #1176, opened 2017-05-21
Removed borderNoise cache.

whichonespink44:
I think I've narrowed down the cause of #1174 to the new borderNoise cache.


whichonespink44:
No longer relevant as of dad0733

PR #1184, opened 2017-05-30
1.10.2/sea level

ThelianTech:
This should have all the latest Changes on them.

### DIFF
--- a/src/main/java/rtg/api/config/RTGConfig.java
+++ b/src/main/java/rtg/api/config/RTGConfig.java
@@ -31,7 +31,7 @@ public class RTGConfig extends Config {
     // Base Terrain Height
     //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    public final ConfigPropertyInt BASE_TERRAIN_HEIGHT;
+    public final ConfigPropertyInt SEA_LEVEL;
 
     //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     // Bedrock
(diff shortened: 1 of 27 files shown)


src/main/java/rtg/api/config/RTGConfig.java
```diff
@@ -286,12 +286,12 @@ public RTGConfig() {
...
 
-        BASE_TERRAIN_HEIGHT = new ConfigPropertyInt(
+        SEA_LEVEL = new ConfigPropertyInt(
                 ConfigProperty.Type.INTEGER,
                 "Sea Level Base Height",
-                "Sea Level",
+                "Base Terrain Height",
                 "The 'Y' Level that the World Spawns at.",
-                80, 52, 132
+                80, 52, 112
```
srs-bsns: The default value should be the vanilla default of 63.

src/main/java/rtg/api/config/RTGConfig.java
```diff
@@ -303,7 +303,7 @@ public RTGConfig() {
             "Number of flat bedrock layers",
             "Bedrock",
             "0 = Normal bedrock (rough pattern); 1-5 = Number of flat bedrock layers to generate",
-            0, 0, 5
+            2, 1, 5
```
srs-bsns, line 306: Why was this changed?

srs-bsns (changes requested):
`Team-RTG:1.10.2/sea-level` should probably be reset to the state of `Team-RTG:1.10.2-dev`, which should remove the conflicts that are occuring.

src/main/java/rtg/api/config/RTGConfig.java
```diff
@@ -286,6 +283,18 @@
...
         //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+        // Base Terrain Height
+        //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+        SEA_LEVEL = new ConfigPropertyInt(
+                ConfigProperty.Type.INTEGER,
+                "Sea Level Base Height",
+                "Base Terrain Height",
+                "The 'Y' Level that the World Spawns at.",
+                80, 52, 112
```
whichonespink44, line 294: We can have the default as 80 whilst we're in development, but this will need to be set to 63 when we merge into dev.

src/main/java/rtg/api/config/RTGConfig.java
whichonespink44: Yeah, let's keep the default at 0.

PR #1185, opened 2017-05-30
Added Polish translation

arutr:
A quick translation of RTG localisation to support Polish language.


whichonespink44:
@Artur96 Thanks for the contribution! 👍

PR #1186, opened 2017-05-30
Merging 1.10.2-dev with 1.11.2-dev

arutr:
I've managed to port commits from `1.10.2-dev` to `1.11.2-dev`. I have experienced some issues with method mismatches, which I've [fix somehow](3d33246). I've also updated the Forge binaries to version 2298.
The build seems stable after flying around for a few minutes and generating several worlds.
If you have any issues then please say so.


srs-bsns:
@Artur96
It's nice of you to take the time to do this, but we are currently focussed on 1.10. We will port to 1.11/1.12 when the next version for 1.10.2 is finished and stable. It is undecided as of yet, but by the time that happens we may skip 1.11 and port directly to 1.12.

In addition to that, you have altered the APIs we use as dependencies. Sorry, but this can not be not allowed.

Issue #1197, opened 2017-06-17
Add better comment for biome configs


whichonespink44:
^

touhouzun:
Less caves and ravines!

srs-bsns:
@touhouzun Your comment has no relevance to this issue.

srs-bsns:
This can be done if the config overhaul is ever completed. Closing this.

PR #1211, opened 2017-07-09
Deco configs

whichonespink44:
This is the candidate PR for the new deco config system. It's still a work-in-progress, but I thought I'd open a PR now in case anyone has any feedback.

When finished, all properties of all decos in all biomes will be configurable. For example, currently boulders can only be enabled/disabled in the biome configs. With the new deco config system, however, users will also be able to configure the boulder deco's `min_y`, `max_y`, `chance`, `strength_factor`, `check_water`, and `boulder_block` properties.

Deco configs are created dynamically and generated during the init() phase.

RTG devs will have the ability to 'restrict' decos and deco properties from being configurable by using a simple flag.

The new deco config system will give users almost complete control over how RTG's biomes are decorated.

Due to the 'technical' nature of some deco properties, verbose comments will be added to each property, explaining what it is and the effect it will have on the deco.

Sensible defaults will be put in place, and upper/lower bounds will be enforced to ensure that users don't inadvertently debauch their worlds.

I've contemplated/experimented with several other approaches (JSON, HOCON, Configurate, etc.), but decided in the end to use a bespoke approach based on the existing biome config system. This is mainly because of the way RTG's deco system has been built. It was either completely re-write the deco system or write a bespoke deco config solution... and I chose the latter. Maybe we can consider a full re-write of the deco system in the next big development sprint, but this will do for now.

[Edit: This PR also contains some general tweaks to the biome config system, and will also include better comments for all biome configs to address #1197.]


srs-bsns:
This branch can be revisited in the future after the port to 1.12. These commits will probably be best added manually by then.

PR #1215, opened 2017-07-16
RTG world by default + proxylicious proxy stuff.

whichonespink44:
This PR makes it so that the RTG world type is selected by default (configurable).

@srs-bsns There's also some proxy refactoring in here, so I won't merge this until the generator settings stuff has been merged  into dev.

### DIFF
--- a/src/main/java/rtg/event/GuiEventHandler.java
+++ b/src/main/java/rtg/event/GuiEventHandler.java
@@ -1,0 +1,60 @@
+package rtg.event;
+
+import net.minecraft.client.gui.GuiCreateWorld;
+import net.minecraft.client.gui.GuiScreen;
+import net.minecraft.world.WorldType;
+
+import net.minecraftforge.client.event.GuiScreenEvent;
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
+import net.minecraftforge.fml.relauncher.Side;
+import net.minecraftforge.fml.relauncher.SideOnly;
+
+import rtg.RTG;
+import rtg.api.util.Accessor;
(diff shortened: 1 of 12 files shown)


src/main/java/rtg/event/GuiEventHandler.java
```diff
@@ -0,0 +1,60 @@
...
+ * @author @Adubbz (https://github.com/Adubbz)
+ *         <p>
+ *         Source: https://github.com/Glitchfiend/BiomesOPlenty/blob/962f1d0db5a8f04d601215ea79944254bcbd48f0/src/main/java/biomesoplenty/common/handler/GuiEventHandler.java
+ *         Modified by: WhichOnesPink (https://github.com/whichonespink44)
+ */
+public class GuiEventHandler {
+
+    public static GuiEventHandler instance = new GuiEventHandler();
+
+    @SideOnly(Side.CLIENT)
```
srs-bsns: This annotation should go on the whole class since it's a client-only class.

whichonespink44: Fixed in fba17c1

Mako11:
Hello, I would like to say why the RTG mode - 1.10.2 is not working on more worlds, why just for the default superstition? I would also need other maps to use this generator maybe this adjustment? or possibly eliminating the blocking of other worlds with the same X-ray generator? I've seen the code

Robijnvogel:
I don't think this is the correct place to contact the developers.
If you want to talk to them you could do that at the Minecraft Forums or [on Discord](https://discord.gg/0wIG7mz3g6TpkBsB).

PR #1237, opened 2018-05-21
Fixed compatibility with latest BoP

Almighty-Alpaca:
This happened due to the Polar Chasm (nether) biome being removed in https://github.com/Glitchfiend/BiomesOPlenty/commit/8db12839d061ea432d6f45c4487130b7cfede9d2.
I've also updated Gradle because the version this project was using is over 2 years old and there have been quite a few performance improvements/bug fixes/new features etc.


srs-bsns:
Sorry, but I'm not accepting PRs to this branch. (It doesn't even currently work.)

I've already known about this issue since the day Forstride removed the biome and with the volatility of the breaking changes he makes whenever he feels like it, and the changes I'll likely be making to RTG's mod compatibility utility, making the changes now is pointless.

Also, pretty much any PR that contains changes to gradle or the build files is an automatic close.
1. There is always the potential that there could be compatibility issues (usually with plugins) when jumping versions that are that far apart and I don't want to have to downgrade the wrapper in the future if I add another plugin. (I know 2.14.1 is old, but there's nothing wrong with it for the time being.)
2. I'd rather test new versions of depenencies before commiting to using them.

whichonespink44:
@Almighty-Alpaca It's not that we don't appreciate the contribution - it's always appreciated when someone makes an effort to get involved, especially with a project that has as limited human resources as RTG has. But yeah, probably not the best time/place for this right now.

As for our policy on PR's that contain changes to Gradle, that's something we should probably put in our coding guidelines, so I'll see what I can do about updating those at some point.

srs-bsns:
I guess I should point out that when dealing with PRs I'm usually just straight forward and to the point, but it shouldn't be taken as a lack of appreciation or anything.
But as Pink said, right now isn't a very good time for outside contributions.

Almighty-Alpaca:
Don't worry, if you already have other plans regarding BoP support i totally understand that. I was just playing around with the current state and noticed the incompatibility.
I really have to disagree regarding Gradle though. It's way more likely to tun into incompatibilities with ancient versions than it is with newer ones. Nobody wants/has the time to fix bug for ancient versions, they will however for newer ones. Staying on the safe side usually means keeping everything up-to-date, especially for such a rapidly evolving tool as Gradle. Otherwise the version gap will just continue to grow indefinitely.

PR #1255, opened 2018-08-11
Re-added support for BOP biomes.

whichonespink44:
Bumped prerelease to beta.1 and bop_ver to 1.12.2-7.0.1.2399


src/main/java/rtg/api/world/deco/DecoPond.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public void generate(IRealisticBiome biome, RTGWorld rtgWorld, Random rand, int worldX, int worldZ, float strength, float river, boolean hasPlacedVillageBlocks) {
+
+        if (this.allowed && rtgConfig.WATER_SURFACE_LAKE_CHANCE.get() > 0) {
+
+            //Surface lakes.
+            for (int i = 0; i < this.loops; i++) {
+
+                int i2 = worldX + rand.nextInt(16);// + 8;
+                int i8 = worldZ + rand.nextInt(16);// + 8;
```
srs-bsns, line 36: You need to uncomment those offsets.

src/main/java/rtg/init/BiomeInit.java
```diff
@@ -227,6 +291,192 @@ private static void init_betteragriculture() {
...
+        }
+        if (BOPBiomes.wetland.isPresent()) {
+            RTGAPI.RTG_BIOMES.addBiomes(new RealisticBiomeBOPWetland());
+        }
+        if (BOPBiomes.woodland.isPresent()) {
+            RTGAPI.RTG_BIOMES.addBiomes(new RealisticBiomeBOPWoodland());
+        }
+        if (BOPBiomes.xeric_shrubland.isPresent()) {
+            RTGAPI.RTG_BIOMES.addBiomes(new RealisticBiomeBOPXericShrubland());
+        }
```
srs-bsns: This is not a safe way to check if BoP biomes should be added.
A list of known `ResourceLocations` should be used and a check made if that ResourceLocation key exists in the biome registry should be made. This will protect against biomes being removed whenever Forstide feels like making breaking changes, which could be at any time.

srs-bsns (commented):
Just to note.. Adding support for the BoP beaches (Origin Beach, and White Beach) is pointless, as they will only ever show up in a world created using BoP's custom GenLayers.
Those GenLayers will only ever be used (are only known to) the BoP BiomeProvider which will only ever be used in a `BIOMESOP` world.

This is the reason I never added support for those beach biomes via the `BeachType` enum.

PR #1269, opened 2018-08-13
Removed volcanoes.

whichonespink44:
Does what it says on the tin!! :))


Ezoteric:
maybe remove your brain while your at it :P

PR #1290, opened 2018-08-19
Complete chinese translation

devkanro:
Hi,
First, thanks a lot for your work, I am a fan of your mod.
Here is my chinese translation for RTG.


whichonespink44:
Hi @higankanshi -

Thanks for your contribution to RTG.

We're still working on the translation keys, so we're not accepting translation PR's at this time since a lot of the translations will need to be updated.

Please feel free to submit another PR after we release our first public Beta, which should be in a couple of weeks. It would be greatly appreciated.

Cheers,
Pink

PR #1337, opened 2019-11-30
Add Galacticraft to invalidBiomes

tie:


Ezoteric:
could also have added MP and EP :P

srs-bsns:
This PR and #1335 were both invalid for 2 reasons.
1. I had already planned for a mod-level blacklist which can be seen in: https://github.com/Team-RTG/Realistic-Terrain-Generation/commit/7b2a4105420e3fbe91f5c00b93615a665a7f22d4
2. The ModCompat$Mods enum ([ModCompat.java#L235](https://github.com/Team-RTG/Realistic-Terrain-Generation/blob/7b2a4105420e3fbe91f5c00b93615a665a7f22d4/src/main/java/rtg/util/ModCompat.java#L235)) is reserved for mods that have biomes that RTG actually supports.

Ezoteric:
nice

PR #1370, opened 2022-04-30
feat: added biome support for Sakura

KisaragiEffective:


srs-bsns:
In any future release of RTG (if that ever happens), manually adding biome support will not be neccessay, making these PRs pointless.

KisaragiEffective:
good to know! I hope that will be released in future!

PR #1376, opened 2024-01-02
BoP Biome updates

Zeno410:
Revised Alps and Mountain foothills to be smaller versions of the full biome

Added Redwood Forest Edge and Snowy Tundra

Corrected RTGBiomeProvider to not adjust rivers in Geographicraft, which has ALWAYS removed all vanilla rivers for RTG.


Zeno410:
I give up on these subordinate branches. I try to push to my branch and it pushes to Team-RTG.

